PR #1, opened 2016-12-08
Use the home environment variable on Plan 9

sirnewton01:
On Plan 9, HOME is home. Also, there was a panic on clone.

### DIFF
--- a/committree.go
+++ b/committree.go
@@ -10,7 +10,11 @@ import (
 )
 
 func (c *Client) GetAuthor() string {
-	configFile, err := os.Open(os.Getenv("HOME") + "/.gitconfig")
+	home := os.Getenv("HOME")
+	if home == "" {
+		home = os.Getenv("home") // On some OSes, it is home
+	}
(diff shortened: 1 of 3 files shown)


committree.go
```diff
@@ -10,7 +10,11 @@ import (
 )
 
 func (c *Client) GetAuthor() string {
-	configFile, err := os.Open(os.Getenv("HOME") + "/.gitconfig")
+	home := os.Getenv("HOME")
+	if home == "" {
+		home = os.Getenv("home") // On some OSes, it is home
+	}
```
driusan: It would probably be better to add a getHomeDir() function which uses build tags to only do the correct thing for the OS that the binary is being built for, but that can be fixed later. (or better yet, use the standard `os.user` package which already does the build tag magic..)

driusan:
I wouldn't decide if a merge commit, squash commit, or rebase merge was least likely to cause problems.. so I went with a merge commit. I guess we'll find out if that was the right choice shortly.

PR #12, opened 2018-04-20
Fixed out of range error in fetch command

themester:


cmd/fetch.go
```diff
@@ -22,6 +22,10 @@ func Fetch(c *git.Client, args []string) {
 	defer file.Close()
 	config := git.ParseConfig(file)
 	repoid := config.GetConfig("remote." + args[0] + ".url")
+	if repoid == "" {
```
driusan: Thanks! But shouldn't this be `len(repoid) < 7` to catch all the cases where the slice in the if statement below might index out of bounds? (or better yet, change the if statement to use `strings.HasPrefix`..)

themester: I am going to change it later. Also I will see all the code and testing it on Plan9. Wait my pull requests hehe

driusan:
Fixed by #31

PR #23, opened 2018-07-10
Add "--template" option for git init

sirnewton01:
Many of the official git tests make use of template parameter to initialize some of the data in the repository before running the test.

### DIFF
--- a/git/init.go
+++ b/git/init.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 )
@@ -10,8 +11,7 @@ type InitOptions struct {
 	Quiet bool
 	Bare  bool
 
-	// Not implemented
-	TemplateDir File
+	Template string
 
 	// Not implemented
 	SeparateGitDir File
@@ -100,6 +100,15 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
 		}
 		fmt.Printf("Initialized empty Git repository in %v/\n", dir)
 	}
+
+                if opts.Template != "" && !filepath.IsAbs(opts.Template) {
+                        wd, err := os.Getwd()
+                        if err != nil {
+                                return c, err
+                        }
+                        opts.Template = filepath.Join(wd, opts.Template)
+                }
+
 	// Now go into the directory and adjust workdir and gitdir so that
 	// tests are in the right place.
 	if err := os.Chdir(c.WorkDir.String()); err != nil {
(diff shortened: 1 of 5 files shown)


git/init.go
```diff
@@ -10,8 +11,7 @@ type InitOptions struct {
 	Quiet bool
 	Bare  bool
 
-	// Not implemented
-	TemplateDir File
+	Template string
```
driusan: The variable rename doesn't bother me, but why change it from File to string? File is a type of string which adds some helper methods for interacting with files. You can always cast it back to a string to interact with the Go standard library, but you lose the compile time benefits of type checking if you don't type it as a File and someone accidentally passes another kind of string.

You can always use `flags.String` and then cast instead of using `flags.StringVar` in the cmd package.

git/init.go
```diff
@@ -111,5 +120,42 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
...
+					return err
+				}
+			}
+
+			return nil
+		})
+
+		if err != nil {
+			return c, err
+		}
```
driusan: I don't think this needs the `err != nil` check. If err is nil, it falls through to the exact same return (and it doesn't look like code that should fall through if anything is ever added below it.)

git/init.go
```diff
@@ -111,5 +120,42 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
...
+		err = filepath.Walk(opts.Template, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+                        path, err = filepath.Rel(opts.Template, path)
+                        if err != nil {
+                                return err
+                        }
+			if info.IsDir() {
+				os.MkdirAll(filepath.Join(c.GitDir.String(), path), 0770)
```
driusan: My review had a comment here that somehow got eaten.. I think this should be 0777. The OS umask takes care of unsetting the bits that aren't supposed to be set (I made a test program that ran `os.MkdirAll("foo", 0777)` and foo came out 0755 to be sure that MkdirAll uses the umask, since the docs don't mention it..)

git/init.go
sirnewton01: Thanks, I think I have addressed your comments with the latest commits. Can you have another look?

driusan:
(Nevermind the conflicts comment, it's only the "Rebase and Merge" option that tells me that.. a merge commit or squash commit doesn't complain on the github frontend..)

sirnewton01:
I'll refactor this into other pull requests and apply the os.File->File fix.

PR #29, opened 2018-07-11
Exit code of 1 when there is an unknown git command

sirnewton01:
When you launch dgit with a subcommand unknown to it it exits with a code of 0 indicating a success. Instead, it should return 1 as the standard git command does.

### DIFF
--- a/main.go
+++ b/main.go
@@ -257,5 +257,6 @@ func main() {
 		}
 	default:
 		fmt.Fprintf(os.Stderr, "Unknown git command %s.\n", subcommand)
+                os.Exit(1)
 	}
 }


main.go
```diff
@@ -257,5 +257,6 @@ func main() {
 		}
 	default:
 		fmt.Fprintf(os.Stderr, "Unknown git command %s.\n", subcommand)
+                os.Exit(1)
```
sirnewton01: This is a different code path than the other one that you had fixed, which was the case where you run dgit with no subcommand at all and it produces the usage. This case involves calling an unrecognized subcommand.

PR #30, opened 2018-07-11
Fix segmentation fault in git config

sirnewton01:
When running git config it sometimes segfaults. This fix avoids the segmentation fault and just reports a message.

### DIFF
--- a/git/config.go
+++ b/git/config.go
@@ -53,8 +53,11 @@ argChecker:
 		g.sections = append(g.sections, section)
 	}
 
-	sec.values[key] = value
-	fmt.Printf("%s", sec)
+        if sec != nil {
+	    sec.values[key] = value
+        } else {
+            fmt.Printf("Couldn't find section %v\n", name)
+        }
 }
 func (g GitConfig) GetConfig(name string) string {
 
@@ -107,6 +110,7 @@ func (s *GitConfigSection) ParseValues(valueslines string) {
 		}
 		varname := strings.TrimSpace(split[0])
 
+                //fmt.Printf("%v\n", varname)
 		s.values[varname] = strings.TrimSpace(strings.Join(split[1:], "="))
 
 	}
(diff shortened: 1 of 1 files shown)


git/config.go
```diff
@@ -53,8 +53,11 @@ argChecker:
 		g.sections = append(g.sections, section)
 	}
 
-	sec.values[key] = value
-	fmt.Printf("%s", sec)
+        if sec != nil {
+	    sec.values[key] = value
+        } else {
```
sirnewton01: Re your previous comment. I checked the official git tool and it seems to create sections on-the-fly when you set a config variable. For now, this just avoid the segfault, which I think is a bit better. To meet parity with the standard git it should auto-create the sections as needed. Btw, it seems that the official git allows as many levels of sections as you want. I'm not sure if I understand the code above that seems to only look for 2/3 levels.

driusan: I think just making the map would both avoid the segfault and make it create the section, but it's been a long time since I've looked at this part of the code and I could be wrong. I'm fine with merging this to avoid the panic for now.

I checked how the official git client encodes deep hierarchies, and it's internally still only encoded as a section/subsection header and key/value pair, it's just that the subsection header has dots in it, so I don't think there's any need to change the internals (though I wouldn't be surprised if there were bugs in dgit in deeply nested config sections.)

ie running:

```
git config --bool foo.bar.baz.quux.qqqq true
```

Resulted in a config file that looks like:

```
[foo "bar.baz.quux"]
    qqqq = true
```

git/config.go
```diff
@@ -107,6 +110,7 @@ func (s *GitConfigSection) ParseValues(valueslines string) {
 		}
 		varname := strings.TrimSpace(split[0])
 
+                //fmt.Printf("%v\n", varname)
 		s.values[varname] = strings.TrimSpace(strings.Join(split[1:], "="))
```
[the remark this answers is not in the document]
driusan: Yes, it looks like I forgot to remove one before committing and didn't notice since I rarely use git config after initially setting up my name/email..

PR #32, opened 2018-07-11
Add template option for git init

sirnewton01:
Let's add the dgit init --template option.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -18,6 +18,7 @@ func Init(c *git.Client, args []string) error {
 
 	flags.BoolVar(&opts.Bare, "bare", false, "Create bare repository")
 	flags.BoolVar(&opts.Quiet, "quiet", false, "Only print errors or warnings")
+	flags.StringVar(&opts.Template, "template", "", "Specify the template directory that will be used")
 	q := flags.Bool("q", false, "Alias of --quiet")
 
 	flags.Parse(args)
--- a/git/init.go
+++ b/git/init.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 )
@@ -10,8 +11,7 @@ type InitOptions struct {
 	Quiet bool
 	Bare  bool
 
-	// Not implemented
-	TemplateDir File
+	Template string
 
 	// Not implemented
 	SeparateGitDir File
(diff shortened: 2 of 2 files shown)


cmd/init.go
```diff
@@ -36,6 +40,17 @@ func Init(c *git.Client, args []string) error {
 		return fmt.Errorf("Invalid init command. Must only provide one directory.")
 	}
 
+	if template != "" {
+		if !filepath.IsAbs(template) {
```
driusan: I'm not sure I see why this needs to be made absolute, but I guess it doesn't hurt either.

git/init.go
```diff
@@ -111,5 +112,38 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
 	}
 	c.WorkDir = WorkDir(wd)
 	c.GitDir = GitDir(wd + "/.git")
+
+	if opts.Template.String() != "" {
```
driusan: You don't need to call the Stringer here since you're comparing to a literal.

git/init.go
```diff
@@ -111,5 +112,38 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
...
+				return err
+			}
+			path, err = filepath.Rel(opts.Template.String(), path)
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				os.MkdirAll(filepath.Join(c.GitDir.String(), path), 0777)
+			} else {
+				newFile, err := os.Create(filepath.Join(c.GitDir.String(), path))
```
driusan: The `git.GitDir` type has a create helper method intended to make this simpler. You could just do

```
newfile, err := c.GitDir.Create(File(path));
```

to create a file named path under the Client's GitDir.

git/init.go
```diff
@@ -111,5 +112,38 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
...
+				if err != nil {
+					return err
+				}
+				defer newFile.Close()
+				f, err := os.Open(filepath.Join(opts.Template.String(), path))
```
driusan: `git.File` (which opts.Template is a type of) also has an Open helper.

git/init.go
```diff
@@ -111,5 +112,38 @@ func Init(c *Client, opts InitOptions, dir string) (*Client, error) {
...
+				defer f.Close()
+				_, err = io.Copy(newFile, f)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		})
+	}
```
driusan: still needs a `return c, err` after the `})` and before the `}`, otherwise the err is being discarded.

(This is the only blocker of my comments, the rest are just to make it a little more idiomatic.)

cmd/init.go
sirnewton01: There is code deeper in the init command that is doing chdir and causing relative paths to not be found relative to where dgit was invoked. That's why the change to absolute paths is needed here.

driusan: Makes sense (well, not really, but it's the chdir that doesn't make sense, your workaround in this PR is fine.. I'll have to look into it sometime.)

sirnewton01:
Thanks @driusan I believe I have addressed most of your points and all of the critical ones in the last round of review.

PR #33, opened 2018-07-11
Fix segmantation fault in git diff

sirnewton01:
There is a segmentation fault in git diff in certain cases exposed by the official test suite. Also, one of the unit tests isn't compiling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: go
+
+go:
+  - "1.10.x"
+  - master
+
+
--- a/git/difffiles.go
+++ b/git/difffiles.go
@@ -53,7 +53,7 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 		}
 		stat, err := f.Stat()
 		if err != nil {
-			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), uint(stat.Size())})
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: go
+
+go:
+  - "1.10.x"
+  - master
```
driusan: This is going to fail because there's two test cases (one in UnidiffApply and one in Status) that are failing and I haven't had a chance to dive into and fix before setting up Travis.

sirnewton01: Would it be ok if I were to comment out those tests for now so that we can hook up travis?

driusan: Weird.. I get intermittent failures on 9front, but pretty consistently those two failures on DragonFly. Now that I'm trying I'm getting the same TestLsFiles and a TestInitRepo failure on a Mac but no Status or Unidiff one. The InitRepo seems to be Mac specific and related to it using weird tmp directories and symlinks. The LsFiles should be looked into and could be a real OS-specific bug on Mac, but doesn't stop Travis.

How about doing something like 

```
if os.Getenv("TRAVIS")=="true" {
    println("Skipping known failure")
    return
}
```

in the status test for now so it can still be run locally?

sirnewton01: That seems to have worked provided that the Travis CI project is set with the environment variable. It was pretty straight forward to set up, but I'm not an admin on dgit so I can't set it up for this project.

driusan: Should be set up now, but I couldn't trigger a build retroactively to verify. We'll see next time.

PR #36, opened 2018-07-11
Add travis icon to readme

sirnewton01:
Let's add the travis CI icon to the readme on the project page so that anyone can see the build health.


driusan:
The whole README probably needs a revamp since there's been a lot of progress too, but no reason not to add this now that it's  set up..

PR #37, opened 2018-07-11
Add logging infrastructure

sirnewton01:
The main function uses the DGIT_TRACE and GIT_TRACE environment variables to decide whether to enable tracing and where the tracing should go. I have converted some of the commented out fmt.Printf statements to use the log package instead.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"github.com/driusan/dgit/cmd"
 	"github.com/driusan/dgit/git"
+        "io/ioutil"
+        "log"
 	"os"
 )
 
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -23,6 +25,32 @@ func requiresGitDir(cmd string) bool {
...
+        log.SetFlags(log.Ldate|log.Ltime|log.Lshortfile)
+        traceLevel := os.Getenv("DGIT_TRACE")
+        if traceLevel == "" {
+           traceLevel = os.Getenv("GIT_TRACE")
+        } 
+      
+        // If no trace level specified then just dump any log output 
+        if traceLevel == "" {
+           log.SetOutput(ioutil.Discard)
+        } else if traceLevel != "1" && traceLevel != "2" {
```
driusan: So is the logic here intended to be that if it's 1 or 2 it goes to stderr, otherwise it's assumed that the var is set to a filename to write to?

sirnewton01: Yes, exactly. I hope that it is reasonable and I believe that it lines up with official git.

driusan: Okay, I just wanted to make sure I know how to use it. Maybe it should be documented somewhere (the new wiki, maybe?)

PR #39, opened 2018-07-12
Add help subcommand

sirnewton01:
Let's add a help subcommand and clean up the help. Also, the standard git outputs usages on stdout, not stderr.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,4 @@ go:
 
 script:
     - diff -u <(echo -n) <(gofmt -d ./)
+    - go test -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,5 @@ go:
   - "1.10.x"
   - master
 
-
+script:
+    - diff -u <(echo -n) <(gofmt -d ./)
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -3,17 +3,17 @@ package cmd
 import (
 	"flag"
 	"fmt"
-        "os"
+	"os"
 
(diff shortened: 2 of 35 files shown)


sirnewton01:
Ah, it's intent based. If you are seeking help then stdout. Otherwise, if it is a usage failure then stderr. Okay, I'll try to see how to adjust the design.

.travis.yml
```diff
@@ -4,4 +4,5 @@ go:
   - "1.10.x"
   - master
 
-
```
sirnewton01: I kept forgetting to format my source code before opening the pull requests. This travis CI addition checks that everything is formatted properly. I can turn this off if you disagree with the philosophy.

driusan: I'm not sure if it'll cause superfluous failures with different go versions, but if it does we can take it out later.

You overrode the default script section though, so it's not running `go test -v ./...` anymore unless you explicitly add it back.

cmd/apply.go
```diff
@@ -3,14 +3,17 @@ package cmd
...
 	"flag"
 	"fmt"
+	"os"
 
 	"github.com/driusan/dgit/git"
 )
 
 func Apply(c *git.Client, args []string) error {
 	flags := flag.NewFlagSet("apply", flag.ExitOnError)
+	flags.SetOutput(os.Stdout)
```
driusan: Why do some of these use `os.Stdout` and other ones use `flag.CommandLine.Output()`?

sirnewton01: This was left over from a previous commit. Thanks, I have fixed those.

PR #46, opened 2018-07-15
Clarify the mnemonics of the command and remove old references to go-git

sirnewton01:


driusan:
It might be worth giving the README a complete overhaul and removing the language trying to scare people off (but still warning them) now that the project has matured a little, too

PR #47, opened 2018-07-15
Add config --list capability

sirnewton01:
This adds the plain --list option to the git config command.

When "-l" is implemented there are other sub-switches that should be added such as showing only the keys.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -43,6 +43,8 @@ func Config(c *git.Client, args []string) {
 	case "--unset":
 		action = "unset"
 		args = args[1:]
+        case "--list":
+                action = "list"
 
 	// Type canonicalization isn't currently supported
 	//  and so we just strip them out and return the raw value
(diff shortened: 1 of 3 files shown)


driusan:
Failing because of your go fmt check.

cmd/config.go
```diff
@@ -114,7 +116,12 @@ func Config(c *git.Client, args []string) {
 		defer outfile.Close()
 		config.WriteFile(outfile)
 		os.Exit(0)
-		return
+        case "list":
+                list := config.GetConfigList()
+                for _, entry := range list {
+                        fmt.Printf("%s\n", entry)
+                }
+                os.Exit(0)
```
driusan: I think this should just return (and the panic's be converted to return an error), this was one of the earlier commands implemented before the package conventions were as well defined.

cmd/config.go
```diff
@@ -79,43 +82,48 @@ func Config(c *git.Client, args []string) {
...
+	case "list":
+		list := config.GetConfigList()
+		for _, entry := range list {
+			fmt.Printf("%s\n", entry)
+		}
+		return nil
 	}
 
-	panic("Unhandled action " + args[0])
+	return errors.New("Unhandled action " + args[0])
```
driusan: It's not a big deal, but the rest of the repo uses fmt.Errorf, not errors.New

PR #48, opened 2018-07-16
Basic git show command

sirnewton01:
The details are in the flags and status.txt. This improve a few of the test cases in the official git test suite.

### DIFF
--- a/cmd/show.go
+++ b/cmd/show.go
@@ -1,0 +1,25 @@
+package cmd
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/driusan/dgit/git"
+)
+
+func Show(c *git.Client, args []string) error {
+	flags := flag.NewFlagSet("show", flag.ExitOnError)
+	flags.SetOutput(flag.CommandLine.Output())
+	flags.Usage = func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "show [options] <commit>...\n")
+		fmt.Fprintf(flag.CommandLine.Output(), "\n\nOptions:\n")
+		flags.PrintDefaults()
+	}
+
+	opts := git.ShowOptions{}
+	flags.StringVar(&opts.Pretty, "pretty", "", "Pretty-print the contents of commit logs in a specified format")
+	flags.Parse(args)
+
+	objects := flags.Args()
+	return git.Show(c, opts, objects)
+}
--- a/git/show.go
+++ b/git/show.go
@@ -1,0 +1,115 @@
+package git
+
+import (
+	"fmt"
+	"strings"
+)
+
+type ShowOptions struct {
+	Pretty string
+}
+
+// Show implementes the "git show" command.
+func Show(c *Client, opts ShowOptions, objects []string) error {
+	if len(objects) < 1 {
+		return fmt.Errorf("Provide at least one commit.")
+	}
+
+	if opts.Pretty != "" && opts.Pretty != "raw" {
+		return fmt.Errorf("Only raw format is supported, not %v", opts.Pretty)
+	}
+
+	commitIds := []CommitID{}
+
+	for _, object := range objects {
+		// Commits only for now
+		commit, err := RevParseCommit(c, &RevParseOptions{}, object)
+		if err != nil {
+			return err
+		}
+
+		commitIds = append(commitIds, commit)
+	}
+
+	for _, commit := range commitIds {
+		if opts.Pretty == "raw" {
+			if err := showCommitRaw(commit, c); err != nil {
+				return err
+			}
+		} else {
+			if err := showCommit(commit, c); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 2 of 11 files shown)


git/show.go
```diff
@@ -0,0 +1,115 @@
+package git
+
+import (
+	"fmt"
+	"strings"
+)
+
+type ShowOptions struct {
+	Pretty string
```
driusan: The man page implies this should have a DiffCommonOptions composed in. I also think "Format" is more descriptive than the Pretty alias.

(Maybe we should have a FormatString type to distinguish it from other types of strings too)

git/show.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+
+	tree, err := cmt.TreeID(c)
+	if err != nil {
+		return err
+	}
+
+	// Headers
+	fmt.Printf("commit %v\ntree %v\nparent %v\nauthor %s %v +0000\ncommiter %s %v +0000\n\n", cmt, tree, parents[0], author, date.Unix(), author, date.Unix())
```
driusan: I think the parent should use string.Join("\n", ...) rather than silently discarding the other parents. ( In fact, I think this whole bit is unnecessary. You can either just call cat-file -p or call c.GetCommitObject and just print the results using the stringer to get the raw commit.)

sirnewton01: Thanks, I'll have a look at doing that for the raw case.

cmd/show.go
```diff
@@ -0,0 +1,26 @@
...
+	flags.SetOutput(flag.CommandLine.Output())
+	flags.Usage = func() {
+		flag.Usage()
+		fmt.Fprintf(flag.CommandLine.Output(), "\n\nOptions:\n")
+		flags.PrintDefaults()
+	}
+
+	opts := git.ShowOptions{}
+	flags.Var(&opts.Format, "format", "Print the contents of commit logs in a specified format")
+	flags.Var(&opts.Format, "pretty", "Alias for --format")
```
driusan: Is this (multiple flags going to the same place) valid or do they overwrite each other?

git/show.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type FormatString struct {
+	value *string
+}
+
+func (f *FormatString) FormatCommit(cmt CommitID, c *Client) (string, error) {
```
driusan: I think client is first everywhere else in the code

cmd/show.go
sirnewton01: They go to the same place. If there is a collision then an error is produced. See git/show.go FormatString type.

driusan: How does the error propagate back to us? Do we need to ensure we check the return of flags.parse?

sirnewton01: I believe that it is handled the same way as if you were to using flags.IntVar() and the user provides a non-integer value for the flag. flags.Parse() will detect it, print the defaults out and os.Exit(2) according to the documentation. We do get an opportunity to provide an error message so that we can tell the user what was wrong with their value. In this case it complains that the format was already set.

sirnewton01: We don't need to check the return of flags.Parse() because we have chosen a flags policy that does an os.Exit(2) before flags.Parse() returns.

git/show.go
```diff
@@ -0,0 +1,117 @@
+package git
+
+import (
+	"fmt"
+	"strings"
+)
+
+type FormatString struct {
```
driusan: Okay, now that I understand what this is doing it explains some mysteries like why the pointer and how we could use flag for multiple options but this seems to be leaking implementation details of the flag package into library code. Can we make this just a type of string in the git package, but have a type in the cmd package to handle the flag interfaces? I think that would be cleaner.

sirnewton01: I'll give this some thought again tomorrow. I see where you're coming from.

Technically, there is no leaking of the flag package here into the git package. It just happens to conform to the interface defined there. Some of those functions may have limited usefulness outside of the context of flags. The choice of the pointer in the struct does seem to make it clear that there is an explicit unset state of nil, which I think is nice in some ways.

Maybe there's a cleaner way to separate the two concerns.

driusan: Using pointers as Maybe types often leads to nil pointer dereferences down the line when someone forgets to check it, and always results in having two incompatible zero values (like on line 13).

I don't think this would be hard to fix. You just need to move your type def, Get, Set, and String methods to the cmd package as-is, then add a `type FormatString string` to the git package which keeps the remaining FormatCommit method  (but it would be simplified and safer by not having to check against nil or dereference the value.)

sirnewton01:
@driusan I think I have found a solution that separates the two concerns fairly nicely. There is a new flags.go file in the cmd package where we can define custom flag value types different dgit specific cases. I have included a string alias value in there that will adjust a string from a pointer and detect if the value is set twice in order to complain. I anticipate that we can add multi-valued string value types in there too and others that we know we will need to address #50

PR #49, opened 2018-07-16
Fix usages, hangs and misinterpreted commands

sirnewton01:
The usages were not quite right for certain subcommands and others that don't use the flags package were hanging and misinterpreting the "--help" flag being passed to them from the help subcommand. This pull request should clean them all up.


driusan:
Failing because of gofmt again

sirnewton01:
Thanks, this newest commit should fix that.

PR #51, opened 2018-07-18
Initial update-index refresh implementation

driusan:
@sirnewton01 I'm sending this as a PR just in case you're going through test cases so you don't duplicate work. This adds an initial implementation of update-index --refresh.

The test case from git's t0000-basic.sh is still failing, so I'm not merging it yet, but I suspect it might be failing because the tests are run sequentially and the previous "validate git-diff-files output against a known state" is failing and needs to work first. (Ie it might be finding a diff because there really is a diff even after refreshing the stat info.)

It also needs dgit test cases written in Go.

### DIFF
--- a/git/difffiles.go
+++ b/git/difffiles.go
@@ -52,7 +52,7 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
 		}
-		stat, err := f.Stat()
+		stat, err := f.Lstat()
 		if err != nil {
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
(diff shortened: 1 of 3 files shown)


driusan:
@sirnewton01 Can you review this? It fixes diff-files enough that the update-index --refresh test is passing (but the previous "validate diff-files against a known state" is still failing, so there's still work to do on diff-files..)

git/difffiles.go
```diff
@@ -52,15 +52,20 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
...
 			continue
 		}
 
 		switch {
 		case stat.Mode().IsDir():
-			fs.FileMode = ModeTree
+			// Since we're diffing files in the index (which only holds files)
+			// against a directory, it means that the file was deleted and
+			// replaced by a directory.
+			//fs.FileMode = ModeTree
```
sirnewton01: Same comment as in the other PR. I'm not a fan of commented out code. The code around can deviate and make it so that it no longer compiles. It affects readability of the code. Git can give us the history so that we can resurrect it if necessary. :)

sirnewton01:
This seems to be superseded by #54. I have reviewed that one too.

driusan:
I just based the other one on this to be sure there were no regressions since they both touched diff-files, but I'll close this anyways

PR #54, opened 2018-07-19
Implement update-index --refresh and make diff-files output match real git client

driusan:
This makes the dgit diff-files command more accurately represent that of git diff-files, where only lstat information appears to be used to look for differences, not the hash of the file content.

Unfortunately, this requires comparing the ctime information, or tests will fail, and Go only easily exposes the mtime in a cross-platform way. I've added a file.CTime() function but only implemented it for DragonFly so far in this PR.

On other systems, we fall back on the old (incompatible, but reasonable to me) behaviour of hashing the file to see if there's differences rather than just `stat`ing them.

### DIFF
--- a/git/difffiles.go
+++ b/git/difffiles.go
@@ -80,9 +80,18 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 			return nil, err
 		}
 		size := stat.Size()
+		ctm, ctmn := f.CTime()
 		log.Printf("%v: Mtime %v idxmtime %v Size: %v idxsize: %v ctime: %v ctimenano:%v\n", f, mtime, idx.Mtime, size, idx.Fsize, idx.Ctime, idx.Ctimenano)
-		if mtime != idx.Mtime || size != int64(idx.Fsize) {
-			hash, _, _ := HashFile("blob", f.String())
+		if mtime != idx.Mtime || size != int64(idx.Fsize) || ctm != idx.Ctime || ctmn != idx.Ctimenano {
+			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), uint(size)})
+			continue
+		}
+
+		// The real git client appears to only compare lstat information and not hash the file. In fact,
+		// if we hash the file then the official git test suite fails on the basic tests. Go, unfortunately,
+		// only exposes mtime in an easy, cross-platform way, not ctime without going through the sys package
+		if idx.Ctime == 0 {
+			hash, _, err := HashFile("blob", f.String())
 
 			if err != nil || hash != idx.Sha1 {
 				val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), uint(size)})
--- a/git/difffiles.go
+++ b/git/difffiles.go
@@ -63,7 +63,6 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 			// Since we're diffing files in the index (which only holds files)
 			// against a directory, it means that the file was deleted and
 			// replaced by a directory.
-			//fs.FileMode = ModeTree
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
 		case !stat.Mode().IsRegular():
--- a/git/difffiles.go
+++ b/git/difffiles.go
@@ -52,7 +52,7 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
 		}
-		stat, err := f.Stat()
+		stat, err := f.Lstat()
 		if err != nil {
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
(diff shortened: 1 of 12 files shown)


driusan:
With this there's only two failing basic tests left:

not ok 77 - update-index D/F conflict
not ok 78 - very long name in the index handled sanely

git/index.go
```diff
@@ -111,6 +144,7 @@ func (d GitDir) ReadIndex() (*Index, error) {
 }
 
 func ReadIndexEntry(file *os.File) (*IndexEntry, error) {
+	log.Println("Reading from index entry from", file.Name())
```
sirnewton01: 👍

git/file_ctime_dragonfly.go
```diff
@@ -0,0 +1,15 @@
+package git
```
sirnewton01: Ouch, I see what you mean. I see that you have implemented dragonfly here. The CI build runs on Linux and there are other OSes that the project aims to support. Do these other OSes need this CTime() implementation too in the near term?

sirnewton01:
That's kind of bizarre that standard git doesn't consult the file hashes and instead looks at the metadata for "diff-files." Other source control systems always fall back on the file content hash.

git/difffiles.go
```diff
@@ -52,15 +52,20 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
 			val = append(val, HashDiff{idx.PathName, idxtree, fs, uint(idx.Fsize), 0})
 			continue
 		}
-		stat, err := f.Stat()
+		stat, err := f.Lstat()
 		if err != nil {
```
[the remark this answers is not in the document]
driusan: I would hope so, since it's supposed to be the same even for non-symbolic links on platforms that *do* support them.. but I just checked the os package, and Plan 9 just calls stat, while windows is slightly more complicated logic which boils down to just calling stat.

git/file_ctime_dragonfly.go
driusan: I implemented DragonFly because I wrote it on DragonFly and didn't want to spend a lot of time giving Linux special treatment since it's mostly just for the test case.

It seems to be failing under Travis (as expected):

```
not ok 70 - validate git diff-files output for a know cache/work tree state
#	
#		cat >expected <<\EOF &&
#	:100644 100644 f87290f8eb2cbbea7857214459a0739927eab154 0000000000000000000000000000000000000000 M	path0
#	:120000 120000 15a98433ae33114b085f3eb3bb03b832b3180a01 0000000000000000000000000000000000000000 M	path0sym
#	:100644 100644 3feff949ed00a62d9f7af97c15cd8a30595e7ac7 0000000000000000000000000000000000000000 M	path2/file2
#	:120000 120000 d8ce161addc5173867a3c3c730924388daedbc38 0000000000000000000000000000000000000000 M	path2/file2sym
#	:100644 100644 0aa34cae68d0878578ad119c86ca2b5ed5b28376 0000000000000000000000000000000000000000 M	path3/file3
#	:120000 120000 8599103969b43aff7e430efea79ca4636466794f 0000000000000000000000000000000000000000 M	path3/file3sym
#	:100644 100644 00fb5908cb97c2564a9783c0c64087333b3b464f 0000000000000000000000000000000000000000 M	path3/subp3/file3
#	:120000 120000 6649a1ebe9e9f1c553b66f5a6e74136a07ccc57c 0000000000000000000000000000000000000000 M	path3/subp3/file3sym
#	EOF
#		git diff-files >current &&
#		test_cmp current expected
#	
ok 71 - git update-index --refresh should succeed
not ok 72 - no diff after checkout and git update-index --refresh
```

I didn't want to add a hack for every OS I'm not using because I'm on the fence about if that's a good idea, or if we should find some other library that exposes ctime more uniformly (but then we're at their mercy as to which OSes are supported).

git/difffiles.go
```diff
@@ -52,15 +52,20 @@ func DiffFiles(c *Client, opt DiffFilesOptions, paths []File) ([]HashDiff, error
...
 			continue
 		}
 
 		switch {
 		case stat.Mode().IsDir():
-			fs.FileMode = ModeTree
+			// Since we're diffing files in the index (which only holds files)
+			// against a directory, it means that the file was deleted and
+			// replaced by a directory.
+			//fs.FileMode = ModeTree
```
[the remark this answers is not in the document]
driusan: agreed, I wasn't certain if this was right (I'm still not), but it's right enough to pass the test cases, so I'll remove the commented out.

git/file_ctime_dragonfly.go
sirnewton01: Yes, the OS requirements are pretty unique for this project, so it is probably best to control this at the cost of some platform-specific code. I say this without seeing how clean/ugly the code is for the other platforms. :)

driusan: Most of them would be pretty similar.  The stat struct would either have a different name,  or slightly different field names, or be in a different order,  but not significantly different.

sirnewton01:
It looks like there was some kind of merge conflict with my last PR that was merged: https://travis-ci.org/driusan/dgit/jobs/406255035#L487

driusan:
I rebased and ran gofmt, I think the revert failing is because of the changes to diff-files and I need to investigate, but shouldn't affect testing #56

sirnewton01:
The latest builds seem to be failing on the TestRevert test. https://travis-ci.org/driusan/dgit/jobs/406270414#L704

driusan:
revert/apply isn't updating the stat info properly, so the changes to diff-files are causing it to detect a dirty tree state. I think I need to implement `git apply --index` to fix that test.

PR #58, opened 2018-07-20
Add var cmd

sirnewton01:
Adds the git var command and the official tests for it to travis.

### DIFF
--- a/cmd/var.go
+++ b/cmd/var.go
@@ -15,6 +15,24 @@ func getLogVar(c *git.Client, config *git.GitConfig, key string, envVar string) 
 		return envVar
 	}
 
+        if key == "GIT_AUTHOR_IDENT" {
+                name := os.Getenv("GIT_AUTHOR_NAME")
+                email := os.Getenv("GIT_AUTHOR_EMAIL")
+                date := os.Getenv("GIT_AUTHOR_DATE")
+                if name != "" && email != "" && date != "" {
+                        return fmt.Sprintf("%s <%s> %s", name, email, date)
+                }
+        }
+
+        if key == "GIT_COMMITTER_IDENT" {
+                name := os.Getenv("GIT_COMMITTER_NAME")
+                email := os.Getenv("GIT_COMMITTER_EMAIL")
+                date := os.Getenv("GIT_COMMITTER_DATE")
+                if name != "" && email != "" && date != "" {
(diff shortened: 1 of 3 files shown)


cmd/var.go
```diff
@@ -0,0 +1,126 @@
...
+		t := time.Now()
+		person := c.GetAuthor(&t)
+		return fmt.Sprintf("%s <%s> %v %s", person.Name, person.Email, time.Now().Unix(), time.Now().Format("-0700"))
+	} else if key == "GIT_EDITOR" {
+		coreEditor, _ := config.GetConfig("core.editor")
+		if coreEditor != "" {
+			return coreEditor
+		}
+
+		visual := os.Getenv("VISUAL")
```
driusan: I think this is missing a check for os.Getenv("GIT_EDITOR") (I'd have to look up the precedence, but I assume before the config)

cmd/var.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		editor := os.Getenv("EDITOR")
+		if editor != "" {
+			return editor
+		}
+
+		return "vi"
+	} else if key == "GIT_PAGER" {
+		corePager, _ := config.GetConfig("core.pager")
```
driusan: This would be missing a GIT_PAGER env check too

cmd/var.go
sirnewton01: It's handled as a catch-all for all vars at the top of the function. The environment variable value that matches the key is passed in to this function.

cmd/var.go
```diff
@@ -0,0 +1,126 @@
...
+		if visual != "" {
+			return visual
+		}
+
+		editor := os.Getenv("EDITOR")
+		if editor != "" {
+			return editor
+		}
+
+		return "vi"
```
driusan: In package git we default to ed, not vi

sirnewton01: Ah, ok. I can change that. I was going on this documentation that seemed to suggest that vi was the usual default. https://git-scm.com/docs/git-var

Ed because it's available on Plan 9?

cmd/var.go
driusan: Ah, I missed that

cmd/var.go
driusan: Partly because it's on Plan 9, partly because ed is the standard unix text editor,  and partly as a joke.

PR #60, opened 2018-07-20
Add file and line number traceability to logs inside tests

sirnewton01:


cmd/main_test.go
```diff
@@ -0,0 +1,13 @@
+package cmd
+
+import (
+	"log"
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
```
driusan: Can we call flag.Parse and not have any output unless -verbose/-v is set?

sirnewton01: Yes, I manage to tie this to that "-v" flag in go test. If you don't specify "-v" then the production code logs are dropped on the floor. Otherwise, they are output along with the verbose testing details.

sirnewton01: The Travis script runs "go test -v" so that we always get verbose output from the CI builds in case that logging is helpful in diagnosing build problems.

PR #61, opened 2018-07-20
Skip the long filename test from the basic suite

sirnewton01:
This one test case from the basic test suite is quite an edge case considering that it would be impractical to create an index entry with such a long name from either Mac or Linux due to local filesystem restrictions. It would be good to get the rest of the basic suite running, but this one can wait until later.

In this PR, there is plumbing in the run-official-tests script to apply patches to the test suites before running them and a patch file that skips this particular test.


driusan:
I'm going to sit on this PR until it's the only thing preventing us from making the official test suite fail the build, just in case there's a solution before then.

But can we move all the test stuff to a subdirectory instead of the root?

sirnewton01:
Okay, sounds good. I have pushed a new commit that restructures all official git test related files into its own subdirectory. Also, travis no longer passes the "-i" parameter to make so any test failures will fail the build. Note that this will likely fail right now until we get the other tests passing.

sirnewton01:
I've pulled out the patch approach and went with an environment variable instead and added that to the travis script. I'm going to check now that it works there.

PR #62, opened 2018-07-20
Fix ls-files so that it doesn't show sibling files

sirnewton01:
The official git test suite was catching a case where when you do a "git ls-files foo" and there is a sibling file called "foobar". It was listing "foobar" even though it is not a child of foo. Resolving this fixes one half of the problem with the t0000-basic.sh 'update-index D/F conflict' test case.

### DIFF
--- a/cmd/lsfiles.go
+++ b/cmd/lsfiles.go
@@ -10,7 +10,7 @@ import (
 // Parses the arguments from git-ls-files as if they were passed on the commandline
 // and calls git.LsFiles
 func LsFiles(c *git.Client, args []string) error {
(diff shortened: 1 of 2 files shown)


sirnewton01:
It seems that the small tweak has broken other areas. I'll need to look at this more in-depth.

sirnewton01:
The problem appears to have been caused by the fact that the HasPrefix() checks were no longer matching exact matches with the extra trailing '/' added to the prefix being matches. This was fixed by also checking for exact string equality in addition to the HasPrefix() match.

PR #64, opened 2018-07-21
Hook replace option in update-index command

sirnewton01:
This PR hooks in replace functionality when specified that will search the index for existing entries based on sha1 hash and sub-paths and remove them.

### DIFF
--- a/git/index.go
+++ b/git/index.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"sort"
+	"strings"
 )
 
 var InvalidIndex error = errors.New("Invalid index")
@@ -187,7 +188,7 @@ const (
 //
 // As a special case, if something is added as Stage0, then Stage1-3 entries
 // will be removed.
-func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stage Stage, size uint32, mtime int64, createEntry bool) error {
+func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stage Stage, size uint32, mtime int64, createEntry bool, replaceEntry bool) error {
 	if stage == Stage0 {
 		defer g.RemoveUnmergedStages(c, path)
 	}
@@ -199,7 +200,26 @@ func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stag
 			entry.Mtime = mtime
 			entry.Fsize = size
 
-			// We found and updated the entry, no need to continue
+			// If replace is true then we search for any entries that
+			//  should be replaced with this one. It could be that there
+			//  are entries with the same SHA-1 hash or there are children
+			//  that are orphaned because of the change from a directory to
+			//  a file.
+			if replaceEntry {
+				newObjects := make([]*IndexEntry, 0, len(g.Objects))
+				for _, e := range g.Objects {
+					if e.Sha1 == s {
+						continue
+					}
+					if strings.HasPrefix(string(e.PathName), string(path)+"/") {
+						continue
+					}
+
+					newObjects = append(newObjects, e)
+				}
+				g.Objects = newObjects
+			}
+
 			return nil
 		}
 	}
@@ -284,7 +304,7 @@ func (g *Index) RemoveUnmergedStages(c *Client, path IndexPath) error {
 // 	else
 // 		add new GitIndexEntry if not found and createEntry is true, error otherwise
 //
-func (g *Index) AddFile(c *Client, file File, createEntry bool) error {
+func (g *Index) AddFile(c *Client, file File, createEntry bool, replaceEntry bool) error {
 	name, err := file.IndexPath(c)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


sirnewton01:
This is breaking some of the dgit tests and so it will need more work but it does in conjunction with the other PR's fix the t0000-basic.sh D/F conflict test case.

git/index.go
```diff
@@ -199,7 +200,26 @@ func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stag
...
 
-			// We found and updated the entry, no need to continue
+			// If replace is true then we search for any entries that
+			//  should be replaced with this one. It could be that there
+			//  are entries with the same SHA-1 hash or there are children
+			//  that are orphaned because of the change from a directory to
+			//  a file.
+			if replaceEntry {
+				newObjects := make([]*IndexEntry, 0, len(g.Objects))
+				for _, e := range g.Objects {
```
driusan: I'm sure this can be better than O(n^2), but for now it's better to get it working slowly than not have it at all.

git/index.go
```diff
@@ -187,7 +188,7 @@ const (
 //
 // As a special case, if something is added as Stage0, then Stage1-3 entries
 // will be removed.
-func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stage Stage, size uint32, mtime int64, createEntry bool) error {
+func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stage Stage, size uint32, mtime int64, createEntry bool, replaceEntry bool) error {
```
driusan: The code path that uses the new param looks pretty self-contained, and only has one caller that sets it, so can we remove the bool param and have a ReplaceStage function instead?

Alternatively, if AddStage and ReplaceStage do share a lot of code, maybe AddStage should take an UpdateIndexOptions to keep it more readable at the call site.

git/index.go
```diff
@@ -199,7 +200,26 @@ func (g *Index) AddStage(c *Client, path IndexPath, mode EntryMode, s Sha1, stag
 			entry.Mtime = mtime
 			entry.Fsize = size
 
-			// We found and updated the entry, no need to continue
+			// If replace is true then we search for any entries that
+			//  should be replaced with this one. It could be that there
+			//  are entries with the same SHA-1 hash or there are children
```
driusan: from the man page, I don't really understand where the bit about the SHA1 comes from

git/index.go
sirnewton01: Right, perhaps we can do performance passes later on or as needed if in real uses it's too slow.

git/index.go
sirnewton01: The arguments list is getting pretty big. Since there's a lot of shared code so far I was thinking of passing in the opts.

git/index.go
sirnewton01: From what I observed from comparing the differences between how official git and dgit handle the D/F conflict test case I saw that the -add --replace was expected to remove one of the entries based on nothing that I could see except for the sha-1 hash of an existing item in the staging area. They shared nothing in common except for the contents. Perhaps there is another way that official git is matching them?

git/index.go
driusan: You can probably replace createEntry with opts.Add if you do that, too. (I'm more concerned with bool params being hard to maintain than the number of params.)

I'm also considering replacing the size/mtime param with some kind of new StatInfo struct in #54, but I'm holding off on that for now because of conflicts with this.

git/index.go
driusan: I don't really know much about how --replace is supposed to work, I'm just going by my understanding of reading the man page.

If it passes the official tests, it's good enough for me.

git/index.go
sirnewton01: @driusan if this is the only thing holding you back from merging can you go ahead and do that? I would like to work on this a little more and validate some assumptions and include your final changes to eliminate some variables. There's a few branches that I'm working off and it's starting to get confusing. :)

git/index.go
sirnewton01: I'm going to spend a bit more time on this and try some more scenarios and re-read the documentation.

git/index.go
driusan: I just merged it without the change and left the FIXME note. It can be cleaned up later, for now it's fixing both diff-files remaining tests. That leaves this and rebasing the other PR to have t0000-basic.sh running.

driusan:
This needs rebase since I merged #54

sirnewton01:
I was wrong about the SHA-1 hash part. It has been removed. It turns out that there can be a parent file or children file entries that might need to be replaced. Also, if replaced is not specified then detecting any of these cases should produce an error. I think it matches the manual page description now. Also, the D/F conflict test case is passing.

git/index.go
```diff
@@ -221,14 +222,44 @@ const (
...
+		g.Objects = newObjects
+
+		return nil
+	}
+
 	// Update the existing stage, if it exists.
 	for _, entry := range g.Objects {
 		if entry.PathName == path && entry.Stage() == stage {
+			err := replaceEntriesCheck()
+			if err != nil {
```
driusan: if you replace this (and the other one) with `if err := replaceEntriesCheck(); err != nil`, I'll merge it.

PR #68, opened 2018-07-22
Include long filename support

driusan:
This adds support for reading long filenames into the index, as well as fixes two other bugs preventing the last t0000-basic test from passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,5 @@ script:
     - diff -u <(echo -n) <(gofmt -d ./)
     - go test -v ./...
     - chmod u+x ./official-git/run-tests.sh
-    - GIT_SKIP_TESTS=t0000.77 ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1009-read-tree-new-index.sh t1304-default-acl.sh t2100-update-cache-badpath.sh t4113-apply-ending.sh t4123-apply-shrink.sh t5705-clone-2gb.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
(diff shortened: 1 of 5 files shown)


git/lsfiles.go
```diff
@@ -2,6 +2,7 @@ package git
 
 import (
 	"io/ioutil"
+	//"fmt"
```
sirnewton01: Unused import. Maybe just remove it?

.travis.yml
```diff
@@ -11,5 +11,5 @@ script:
     - diff -u <(echo -n) <(gofmt -d ./)
     - go test -v ./...
     - chmod u+x ./official-git/run-tests.sh
-    - GIT_SKIP_TESTS=t0000.77 ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1009-read-tree-new-index.sh t1304-default-acl.sh t2100-update-cache-badpath.sh t4113-apply-ending.sh t4123-apply-shrink.sh t5705-clone-2gb.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
```
sirnewton01: 👍

git/updateindex.go
```diff
@@ -129,9 +129,32 @@ func UpdateIndexFromReader(c *Client, opts UpdateIndexOptions, r io.Reader) (*In
 			return nil, fmt.Errorf("update-index --index-info variant 1 not implemented")
 		case 3:
 			switch len(spaces[1]) {
-			case 20:
```
sirnewton01: What happened to case 20? Should it produce an error about variant 3?

driusan: It was always wrong.  A Sha1 is 20 bytes long,  which is 40 characters when written as a hex string. It just wasn't caught before because modes are never 20 characters either.

git/file.go
```diff
@@ -133,3 +134,18 @@ func (f File) Remove() error {
 func (f File) Open() (*os.File, error) {
 	return os.Open(f.String())
 }
+
```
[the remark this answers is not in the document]
driusan: Good catch. I'll try and fix it tomorrow. We'll still need a solution for ** globs, though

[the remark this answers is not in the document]
sirnewton01: Right, but hopefully that and the trailing '/' for directories things can be captured as special cases.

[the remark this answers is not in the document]
driusan: Fixed. I amended the commits because I'm trying to keep this cleanly separated into 3 atomic commits.

git/lsfiles.go
driusan: fixed

git/file.go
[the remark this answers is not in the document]
sirnewton01: 👍

PR #69, opened 2018-07-22
Fix dgit clone (#40)

b1narykid:
Clone command was broken when #40 removed `--set` switch.

### DIFF
--- a/cmd/clone.go
+++ b/cmd/clone.go
@@ -67,7 +67,7 @@ func Clone(c *git.Client, args []string) error {
 	}
 
 	if err := Config(c, []string{"remote.origin.url", repoid}); err != nil {
-		return nil
+		return err
 	}
 	if err := Config(c, []string{"branch.master.remote", "origin"}); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


sirnewton01:
Thanks, good catch on this one. The automated testing didn't find it.

cmd/clone.go
```diff
@@ -66,12 +66,12 @@ func Clone(c *git.Client, args []string) error {
 		return err
 	}
 
-	Config(c, []string{"--set", "remote.origin.url", repoid})
```
sirnewton01: I think it would be good to check the errors coming back from these Config calls even though that wouldn't have fixed this particular problem, there could be other error cases that are missed.

b1narykid: No problem.  What about `Fetch` and `Reset` calls below?

driusan:
Ouch. We should find a way to have better automated tests for something this important.

cmd/clone.go
driusan: Fetch doesn't return an error,  Reset should probably be checked

driusan: I committed it and then amended to add the reset check.

PR #71, opened 2018-07-23
Add basic ignore functionality

sirnewton01:
Adds a new check-ignore command. The status command uses the check-ignore backing logic to filter out any untracked changes. The official git t0008-ignores.sh suite went from completely failing to 297/392 passing.


git/status.go
```diff
@@ -346,6 +346,11 @@ func StatusLong(c *Client, files []File, untracked StatusUntrackedMode, linepref
 				if err != nil {
 					return "", err
 				}
+				// TODO instead of filtering afterwards, perhaps the untracked should be filtered at the source
+				matches, _ := CheckIgnore(c, CheckIgnoreOptions{}, []File{fname})
```
driusan: Why don't we put this check in LsFiles and include ExcludeStandard in the opts? That would go a lot farther than anything a filter in every porcelain command

sirnewton01:
The latest commits add more capabilities with the ignore patterns. Next step is to integrate the ignores into ls-files so that other plumbing commands get the benefit, such as status and add.

sirnewton01:
Thanks @driusan I see that you anticipated that the LsFiles() function would be the place where ignore capabilities are centralized. Hooking it into status/add was really straight forward. With the latest changes in this PR the t0008-ignores.sh official test has only 87 failures out of 392. I'm going to try to whittle this number down further so that I can add it to the travis CI.

official-git/fix-ignores-test.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/t/t0008-ignores.sh b/t/t0008-ignores.sh
+index 89544dd83..217068733 100755
+--- a/t/t0008-ignores.sh
++++ b/t/t0008-ignores.sh
+@@ -187,7 +187,7 @@ test_expect_success 'setup' '
+ 		git init &&
+ 		echo a >a &&
+ 		git add a &&
+-		git commit -m"commit in submodule"
```
sirnewton01: There's no space between the -m and the commit message, which causes go's flags package to fail on it and also the git manual specifies a space between -m and the message. It seems that the official git command-line just supports this unspecified syntax and the test here happens to use it.

git/checkignore.go
```diff
@@ -0,0 +1,244 @@
...
+			return nil, fmt.Errorf("Path %v is not in the git work directory.", path.String())
+		}
+
+		dir := filepath.Dir(abs)
+
+		for {
+			gitignore := filepath.Join(dir, ".gitignore")
+			log.Printf("Checking .gitignore in %s\n", gitignore)
+			ignorePath, err := filepath.Rel(dir, abs)
+			pattern, lineNumber, err := findPatternInIgnoreFile(c, gitignore, ignorePath, isDir)
```
driusan: What about matches that come from .gitignores higher up in the tree? I think the logic for directories is easier to reason about recursively than iteratively. For instance, we could parse the ignore file and return a slice of patterns, then pass that slice to children to deal with parent ignores. (Which also has the advantages of only parsing the file once). If that was done and the ls-files inversion, it would be easy to add exclude-per-directory and exclude-from

git/checkignore.go
```diff
@@ -0,0 +1,244 @@
...
+
+		if isEof {
+			break
+		}
+	}
+
+	return "", 0, nil
+}
+
+func matchesGlob(path string, isDir bool, pattern string) bool {
```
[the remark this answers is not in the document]
sirnewton01: For now, I was going to keep this local because it doesn't quite fit the File.MatchGlob pattern after all. For example, this ones needs to know if the path points to a directory or not for some of the matching logic, but the path may be relative to arbitrary positions within the work directory.

git/checkignore.go
sirnewton01: Yes, I could think about this a bit more. Some of the complexity comes from the fact that in order to match a pattern in a gitignore file you need to make the path of the file to be ignore/not ignored relative to that gitignore file in order to check the matches.

git/checkignore.go
```diff
@@ -0,0 +1,244 @@
...
+
+			if pattern != "" {
+				patternMatches[idx] = IgnoreMatch{Pattern: pattern, LineNum: lineNumber, Source: File(".git/info/exclude"), PathName: path} // TODO .git/info/exclude is hard coded here instead of being calculated from c.GitDir
+			}
+		}
+
+		// Be sure to assign the pathname in all cases so that clients can match the outputs to their inputs
+		patternMatches[idx].PathName = path
+
+		// TODO: consider the other places where ignores can come from, such as core.excludesFile and .git/info/exclude
```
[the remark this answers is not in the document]
driusan: I'm not an expert on the exclude semantics, but from how I interpret the man page it just affects whether it calls `ls-files -o` or `ls-files -c -o`

git/ignore_test.go
```diff
@@ -0,0 +1,117 @@
...
+		t.Fail()
+	}
+
+	nonZero := IgnoreMatch{PathName: File(".././abc/xyz/foo.txt"), IgnorePattern: IgnorePattern{Pattern: "/**/*.txt", LineNum: 5, Source: File("xyz/.gitignore")}}
+	if nonZero.String() != fmt.Sprintf("xyz/.gitignore:5:/**/*.txt\t.././abc/xyz/foo.txt") {
+		t.Fail()
+	}
+}
+
+func TestIgnorePatternMatches(t *testing.T) {
```
driusan: This test seems to want to be a table driven test

git/ignore.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	file, err := patternFile.Open()
+	if err != nil {
+		return []IgnorePattern{}, err
+	}
+	defer file.Close()
+
+	ignorePatterns := []IgnorePattern{}
+	reader := bufio.NewReader(file)
```
driusan: I haven't looked into this loop in detail, but from a cursory glance it seems that this might be easier using a bufio.Scanner to scan lines instead of a bufio.Reader

git/ignore.go
```diff
@@ -0,0 +1,275 @@
...
+	alreadyParsed := make(map[string]bool)
+	finalPatterns := []IgnorePattern{}
+
+	for _, path := range paths {
+		abs, err := filepath.Abs(path.String())
+		if err != nil {
+			return nil, err
+		}
+
+		// Let's check that this path is in the git work dir first
```
driusan: This check seems useful enough that it might make sense to make it a method on File rather than inline in the loop.

git/file.go
```diff
@@ -72,6 +72,45 @@ func (f File) IsDir() bool {
 
 }
 
+// Returns true if the file is inside a submodule and the submodule file.
+func (f File) IsInSubmodule(c *Client) (bool, File, error) {
```
driusan: Can you add a FIXME saying that when we implement submodules this should be inverted? (Submodules contain Files, not the other way around,  but we don't currently have a submodule type..)

cmd/checkignore.go
```diff
@@ -0,0 +1,234 @@
...
+				fmt.Fprintf(os.Stderr, "fatal: pathspec '%v' is beyond a symbolic link\n", p)
+				os.Exit(128)
+			}
+
+			paths = append(paths, git.File(p))
+		}
+
+		// Invoke the ignore routines directly, rather than relying on
+		//  LsFiles because we need to be able to do things such as
+		//  return non-matches and match details that aren't supported there.
```
driusan: I don't think I understand this comment.  Isn't determining whether nonmatches or matches are returned what the `-i` option (which admittedly isn't currently implemented, but that's effectively what this PR is doing) toggles? What details that aren't supported need to be matched?

sirnewton01:
@driusan I've reorganized things here quite a bit hoping that it is a much cleaner implementation with better separation of concerns. I've gotten the number of official git tests down to only 28 out of 395 for the ignore tests. The remainder are mostly related to the fact that I have not yet implemented the negation '!' operator and a couple of edge cases with escaping of whitespace vs. the slash character itself.

git/ignore.go
sirnewton01: I wish that were the case, but unfortunately buffo.Scanner will miss the last line of a stream if there is no newline character at the end. :( This is why there is the pain of having to use the reader directly.

git/file.go
sirnewton01: +1

cmd/checkignore.go
sirnewton01: Right the "-i" option is not yet enabled and the LsFiles won't return back the pattern that matched or wasn't matched so that checkignore can output that with the -v flag.

sirnewton01:
I have added all of the passing official git ignore tests to travis to catch regressions. The majority of the failing tests are because of the negation operator, which I think should be tackled as a separate issue.

Issue #70, opened 2018-07-22
Clone needs refactoring


driusan:
The cmd/Clone function predates the repo being converted into a cmd and a git package,  and directly invokes other commands in the cmd package.  It should be refactored to only use the git package, so that we can improve the test coverage of it. 

(There are a few other commands in this boat,  but  #69 just reminded me of this one. )

sirnewton01:
Unfortunately, the currently functional official git tests didn't catch this. It's possible that they were being masked by other problems. It will be good to improve local test coverage for this blind spot.

PR #74, opened 2018-07-29
Clean up clone and fetch commands, add convenience for loading/writing configs

sirnewton01:


sirnewton01:
This is my initial cleanup on this. I hope to write some automated tests to catch regressions like the one that led to #70

git/config.go
```diff
@@ -239,5 +243,68 @@ func ParseConfig(configFile io.Reader) GitConfig {
...
+	home := os.Getenv("HOME")
+	if home == "" {
+		home = os.Getenv("home") // On some OSes, it is home
+	}
+
+	if home == "" {
+		return GitConfig{}, fmt.Errorf("Global git configuration could not be found since HOME and home environment variables were not defined.")
+	}
+
+	fname := home + "/.gitconfig"
```
driusan: I think everything above this should be replaced with a private function to get the global config which can be modified with build tags for different OSes, then we could use,  for instance $home/lib/gitconfig where it sense. (Also, I suspect it's only a matter of time before someone asks for xdg config support)

git/config.go
```diff
@@ -239,5 +243,68 @@ func ParseConfig(configFile io.Reader) GitConfig {
...
+	err := os.Remove(g.fname)
+	if err != nil {
+		return err
+	}
+	configFile, err := os.Create(g.fname)
+	if err != nil {
+		return err
+	}
+	g.WriteFile(configFile)
+	err = configFile.Close()
```
driusan: This could just be "return configFile.Close()".

git/fetch.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+type FetchOptions struct {
+}
+
+func FetchRepository(c *Client, opts FetchOptions, repository string) error {
```
driusan: Why isn't this just Fetch to mirror other commands where it has the same name as the command?

git/clone_test.go
```diff
@@ -0,0 +1,93 @@
+package git
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestCloneRepo(t *testing.T) {
```
driusan: I'm not crazy about test cases needing network access to run.  Is there any way we could make the fetching use a fake network connection? Maybe a better way to ensure clone doesn't break in the real world is to just use dgit to clone something arbitrary under travis

git/config.go
sirnewton01: 👍

git/config.go
sirnewton01: 👍

git/fetch.go
sirnewton01: I was thinking that when we support more flavours of fetch that there might be different arguments to the function such as the group, or even fetch with no repository at all. There's one variant that allows you to specify more than one repository. Some of those don't really make sense in the options. It's a little unfortunate that go doesn't support function overloading.

I will rename this back to match the rest of the commands and defer some of the decision making until later.

git/clone_test.go
sirnewton01: Agreed, but I was thinking that this would cover the primary purpose of dgit, which to allow it to "go get" something where GitHub is by far the most popular place for people to "go get" stuff. I'm not sure how far we would want to go to mock out an https or ssh connection. Do you have something in mind for this?

driusan: No, that's why I was thinking maybe just cloning something in Travis instead of a Go test might be a good idea.

git/fetch.go
driusan: We can always refactor later (in fact,  we'll almost certainly need to once fetch-pack is implemented)

git/clone_test.go
sirnewton01: :+1 Ah, I understand now. Okay, I'll try hooking that up.

.travis.yml
```diff
@@ -12,4 +12,5 @@ script:
     - go test -v ./...
     - chmod u+x ./official-git/run-tests.sh
     - GIT_SKIP_TESTS='t0008.30[5-9] t0008.310 t0008.321 t0008.323 t0008.37[0-9] t0008.38[0-7] t0008.39[1-2]' ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0008-ignores.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1009-read-tree-new-index.sh t1304-default-acl.sh t2100-update-cache-badpath.sh t4113-apply-ending.sh t4123-apply-shrink.sh t5705-clone-2gb.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
+    - ./dgit clone https://github.com/driusan/dgit.git /tmp/dgit.$$
```
driusan: I don't understand how this works without having done a go build first,  but it seems to actually be running in the travis logs, so I'll take it

PR #76, opened 2018-08-02
Fix all basic 3-way merge tests

driusan:
This fixes the basic 3-way merge tests from the official git test suite.

Primarily, this involved handling the case where something is a directory in one stage and a file in a different stage, as well as adding tests that the index and work tree matches either stage 2 (or the stage that it's going to collapse to.)

### DIFF
--- a/git/readtree.go
+++ b/git/readtree.go
@@ -124,10 +124,68 @@ func ReadTreeMerge(c *Client, opt ReadTreeOptions, stage1, stage2, stage3 Treeis
 	}
 	var dirs []IndexPath
 
-	// Checking for merge conflict with index
+	// Checking for merge conflict with index. If this seems like a confusing mess, it's mostly
+	// because it was written to pass the t1000-read-tree-m-3way test case from the official git
+	// test suite.
+	//
+	// The logic can probably be cleaned up.
 	for path, orig := range origMap {
-		ours, ok := ours[path]
-		if !ok || ours.Sha1 != orig.Sha1 {
+		o, ok := ours[path]
+		if !ok {
+			// If it's been added to the index in the same state as Stage 3, and it's not in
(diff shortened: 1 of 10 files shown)


git/readtree.go
```diff
@@ -92,43 +100,183 @@ func ReadTreeMerge(c *Client, opt ReadTreeOptions, stage1, stage2, stage3 Treeis
...
+		}
+
+		// Handle the subfiles in any directory that had a conflict
+		// by just adding them in the appropriate stage, because
+		// there's no way for a directory and file to not be in
+		// conflict.
+		for _, d := range dirs {
+			if strings.HasPrefix(string(path), string(d+"/")) {
+				if p, ok := base[path]; ok {
+					if err := idx.AddStage(c, path, p.Mode, p.Sha1, Stage1, p.Fsize, time.Now().UnixNano(), UpdateIndexOptions{Add: true}); err != nil {
```
sirnewton01: Should this panic or should it return the error to the caller?

sirnewton01:
It looks good to me. I have just one comment about panics.

Sometimes the official tests check things that aren't well spelled out by the documentation, which makes it challenging with re-implementing it like we're doing here. It's a risk with using the official git tests to verify our implementation. The code is sometimes weird because we are trying to make the tests pass, but the payoff is that we can unlock more parts of their test suite that verify the correctness of things that are well described in the documentation, but perhaps not implemented correctly on our end.

git/readtree.go
driusan: I couldn't decide that when I was writing it. We *could* propagate the error, but  If read-tree can't add a stage to the index, something has gone horribly, horribly wrong and it probably needs investigating.

driusan: I changed it to return an error, I'll merge it once it passes Travis.

PR #77, opened 2018-08-02
Add --force-remove option to update-index

driusan:
This allows the setup of the two way read-tree test to succeed.

### DIFF
--- a/cmd/updateindex.go
+++ b/cmd/updateindex.go
@@ -203,7 +203,7 @@ func UpdateIndex(c *git.Client, args []string) error {
 		}
 	}
 
-	if len(files) > 0 {
+	if len(files) > 0 || opts.Refresh || opts.ReallyRefresh || opts.IndexInfo != nil {
 		idx, err = git.UpdateIndex(c, idx, opts, files)
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


cmd/updateindex.go
```diff
@@ -156,20 +156,58 @@ func UpdateIndex(c *git.Client, args []string) error {
...
 	// Load the index file and call UpdateIndex on it.
 	idx, err := c.GitDir.ReadIndex()
 	if err != nil {
 		return err
 	}
 
-	newidx, err := git.UpdateIndex(c, idx, opts, files)
-	if err != nil {
-		return err
+	for _, val := range vals {
```
sirnewton01: I can see that we're bumping into a limitation of flags here that I don't have an idea how to work around. Because they are not declared as flags they won't show up in usage. You can probably put fmt.Fprintf() statements at the end of the flags.Usage override at the top of this function so that they would show up in the help. Also, I'm not sure if this affects the subcommandUsage in the main.go case statement for this subcommand. It may be worth having a quick look and see.

driusan: They do show up in the usage and are declared as flags though, the problem is when there's filenames separating them and they become context sensitive. Ie in "--add A B --force-remove C" flags will stop parsing once it sees the "A", because it's encountered something that isn't a flag, even though something later ("--force-remove") is a known flag which modifies the things that come after it, by that point it's already stopped parsing.

PR #81, opened 2018-08-03
Fix fast-forward read-tree tests

driusan:
This fixes most of the read-tree 2-way (fast-forward) tests from the official git test suite.


cmd/commondiff.go
```diff
@@ -22,6 +22,7 @@ func parseCommonDiffFlags(c *git.Client, options *git.DiffCommonOptions, default
 	s := flags.Bool("s", false, "Alias of --no-patch")
 	unified := flags.Int("unified", 3, "Generate <n> lines of context")
 	U := flags.Int("U", 3, "Alias of --unified")
+	U0 := flags.Bool("U0", false, "Alias of -U 0. (This is primarily for test compatibility)")
```
sirnewton01: Ah, here is the -Ux flag.

PR #84, opened 2018-08-05
Add ls-files --no-empty directory and fix file sorting order

driusan:
Go string comparison doesn't produce the same output as git (which
uses memcmp order) for sorting files, so this updates the sort.ByPath
to emulate memcmp in order to make it easier to compare test output
between git and dgit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,7 @@ script:
     - diff -u <(echo -n) <(gofmt -d ./)
     - go test -v ./...
     - chmod u+x ./official-git/run-tests.sh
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -11,5 +11,7 @@ script:
     - diff -u <(echo -n) <(gofmt -d ./)
     - go test -v ./...
     - chmod u+x ./official-git/run-tests.sh
-    - GIT_SKIP_TESTS='t0008.30[5-9] t0008.310 t0008.321 t0008.323 t0008.37[0-9] t0008.38[0-7] t0008.39[1-2]' ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0008-ignores.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1000-read-tree-m-3way.sh t1001-read-tree-m-2way.sh t1002-read-tree-m-u-2way.sh t1009-read-tree-new-index.sh t1304-default-acl.sh t2100-update-cache-badpath.sh t3002-ls-files-dashpath.sh t3006-ls-files-long.sh t4113-apply-ending.sh t4123-apply-shrink.sh t5705-clone-2gb.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
+    # t0008 tests that are skipped require ! to negate a pattern. (GitHub issue #72)
```
sirnewton01: The negation one seems to be coming up more and more. I wonder if it should be tackled soon?

driusan: There aren't any new cases of it, I just added a comment to explain the ones that were already there.

PR #90, opened 2018-08-08
Improve cat-file enough that commit-tree tests pass

driusan:
This improves the handling of git cat-file enough that the
commit-tree test (which depends on it) succeeds.

cat-file is still missing enough options that the cat-file
tests themselves fail (in particular, it does a lot of --batch
tests and annotated tags which we don't yet support.)

### DIFF
--- a/cmd/catfile.go
+++ b/cmd/catfile.go
@@ -32,7 +32,6 @@ func CatFile(c *git.Client, args []string) error {
 		flags.Usage()
 		return nil
 	case 1:
-		println("I am here")
 		shas, err := git.RevParse(c, git.RevParseOptions{}, oargs)
 		if err != nil {
 			return err


cmd/catfile.go
```diff
@@ -22,19 +22,41 @@ func CatFile(c *git.Client, args []string) error {
...
-		return err
-	}
-	for _, s := range shas {
-		val, err := git.CatFile(c, s.Id, options)
+	switch len(oargs) {
+	case 0:
+		flags.Usage()
+		return nil
+	case 1:
+		println("I am here")
```
sirnewton01: Leftover from testing?

driusan: looks like it, I'll fix it tonight

Issue #92, opened 2018-08-10
Compile errors on darwin


sirnewton01:
There appear to be some compile problems with darwin (Mac) in the git/file_inode_darwin.go. These changes seem to be working for me now.

```
func (f File) INode() uint32 {
        stat, err := f.Lstat()
        if err != nil {
                return 0
        }
        rawstat := stat.Sys().(*syscall.Stat_t)
        return uint32(rawstat.Ino)
}
```

PR #93, opened 2018-08-11
Fix typo in darwin inode file

driusan:


driusan:
@sirnewton01 can you verify that I didn't miss something else and this fixes #92?

git/file_inode_darwin.go
```diff
@@ -10,5 +10,5 @@ func (f File) INode() uint32 {
 		return 0, 0
```
sirnewton01: The return is only a single value here.

PR #98, opened 2018-08-12
Go get tests

sirnewton01:


go-get-tests.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "Go get a package"
+go get ${TEST_PKG} || echo "Known problem with go get, it invokes the submodule command"
+test -d ${TEST_GIT_DIR} || (echo "ERROR: Go get didn't work"; exit 1)
+
+test -f $DGIT_TRACE || (echo "ERROR: Dgit wasn't called for the go get test"; exit 1)
+unset DGIT_TRACE
+
+echo "Reset the package back one commit from master"
+$ORIG_GIT -C ${TEST_GIT_DIR} reset HEAD^1
+$ORIG_GIT -C ${TEST_GIT_DIR} checkout .
```
driusan: I don't get it. What is this for?

.travis.yml
```diff
@@ -13,8 +13,9 @@ script:
     - pwd
     - diff -u <(echo -n) <(gofmt -d ./)
     - go test -v ./...
+    - chmod u+x ./go-get-tests.sh
```
driusan: Why not just check it into the repo as +x?

.travis.yml
```diff
@@ -13,8 +13,9 @@ script:
...
+    - ./go-get-tests.sh
     - chmod u+x ./official-git/run-tests.sh
     # t0008 tests that are skipped require ! to negate a pattern. (GitHub issue #72)
     # t3000.7 requires git pack-refs
     - GIT_SKIP_TESTS='t0008.30[5-9] t0008.310 t0008.321 t0008.323 t0008.37[0-9] t0008.38[0-7] t0008.39[1-2] t3000.7' ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0008-ignores.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1000-read-tree-m-3way.sh t1001-read-tree-m-2way.sh t1002-read-tree-m-u-2way.sh t1003-read-tree-prefix.sh t1009-read-tree-new-index.sh t1100-commit-tree-options.sh t1304-default-acl.sh t2000-checkout-cache-clash.sh t2001-checkout-cache-clash.sh t2002-checkout-cache-u.sh t2009-checkout-statinfo.sh t2100-update-cache-badpath.sh t3000-ls-files-others.sh t3002-ls-files-dashpath.sh t3006-ls-files-long.sh t3020-ls-files-error-unmatch.sh t4113-apply-ending.sh t4123-apply-shrink.sh t5705-clone-2gb.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
-    - ./dgit clone https://github.com/driusan/dgit.git /tmp/dgit.$$
```
driusan: Is this removed because it's implicitly tested by `go get`?

go-get-tests.sh
sirnewton01: It's to force it back one commit from origin/master so that pull can bring it back up to date. I used the official git to do this because I don't think we have the rev-parsing capability to handle the HEAD^1 (HEAD, backward one commit).

.travis.yml
sirnewton01: Right, I felt it was redundant because of the go get, which does the clone and more.

.travis.yml
sirnewton01: It is right now, but you never know if a windows commit comes in that backs it out. Just being a bit defensive.

go-get-tests.sh
driusan: I see. Makes sense. I'll merge this, and if there's any problems caused we can always revert it.

PR #100, opened 2018-08-14
Various improvements to git read-tree

driusan:
This makes various improvements to git read-tree test t1004. It enforces the --exclude-per-directory command line restrictions (can only pass once, must include -u), adds the ability to checkout tags, and fixes a bug with 'git checkout -b foo bar' where it would create foo at commit bar, but not change the HEAD to foo.

(This leaves 6 failing tests in t1004)


git/readtree.go
```diff
@@ -560,8 +560,6 @@ func checkMergeAndUpdate(c *Client, opt ReadTreeOptions, origidx map[IndexPath]*
 	}
 
 	// Keep a list of index entries to be updated by CheckoutIndex.
-	//files := make([]File, 0, len(newidx.Objects))
```
sirnewton01: It's good to see the removal of dead code. 👍

git/checkout.go
```diff
@@ -22,10 +22,8 @@ type CheckoutOptions struct {
 	// Not implemented
 	Stage Stage
 
-	// Not implemented
-	Branch string // -b
-	// Not implemented
-	ForceBranch bool // use branch as -B
+	Branch      string // -b
+	ForceBranch bool   // use branch as -B
```
sirnewton01: I don't see where ForceBranch is being used. Did you want to preserve the // Not implemented comment for it until it is used?

driusan: Sure, I added the comment back.

PR #104, opened 2018-08-15
Add special case for merge-base

driusan:
Add special case for a merge-base containing only
two commits, where if one commit is an ancestor of
the other that is taken as the merge-base without
doing a search. This makes `dgit merge-base A B`
and `dgit merge-base B A` return the same commit in
these cases (which is what real git does), rather
than returning the zero commit in one but the right
commit in the other.

### DIFF
--- a/git/mergebase.go
+++ b/git/mergebase.go
@@ -9,6 +9,23 @@ type MergeBaseOptions struct {
 }
 
 func MergeBase(c *Client, options MergeBaseOptions, commits []Commitish) (CommitID, error) {
+	if len(commits) == 2 {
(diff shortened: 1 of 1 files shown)


git/mergebase.go
```diff
@@ -9,6 +9,23 @@ type MergeBaseOptions struct {
 }
 
 func MergeBase(c *Client, options MergeBaseOptions, commits []Commitish) (CommitID, error) {
+	if len(commits) == 2 {
```
sirnewton01: This looks good to me. Did you want to include a test case for this since it looks like it could be reproduced with a simple repo/branch with two commits in it?

PR #106, opened 2018-08-16
Implement pull

sirnewton01:
I have implemented a simple pull out of fetch and merge using only the options available to those other subcommands at the moment.


cmd/pull.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return err
+	}
+
+	if flags.NArg() < 1 {
+		// TODO simplistic, probably won't work in all cases
+		head := c.GetHeadBranch().BranchName()
+		branchremote, _ := config.GetConfig("branch." + head + ".remote")
+		repository = branchremote
+		refspec = []string{repository + "/" + head}
```
driusan: 1. my .git/config seems to have store the refspec in `branch.$BRANCHNAME.merge`
2. I don't think what you're storing in the refspec variable is an actual refspec. (refspecs are things that live under the refs/ subdirectory of $GIT_DIR.)

cmd/pull.go
```diff
@@ -0,0 +1,82 @@
...
+	} else if flags.NArg() == 2 {
+		repository = flag.Arg(0)
+		refspec = flag.Args()[1:]
+	} else {
+		flags.Usage()
+		os.Exit(1)
+	}
+
+	err = git.Fetch(c, fetchopts, repository)
```
driusan: I think everything below here belongs in a Pull func in the git package, not in the cmd package.

cmd/pull.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return err
+	}
+
+	others := make([]git.Commitish, 0, len(refspec))
+	for _, name := range refspec {
```
driusan: I don't see any way for there to ever be more than 1 value in refspec, so I'm not sure why this is a loop. (or why you're using a slice.)

cmd/pull.go
sirnewton01: There's some very mysterious stuff happening in there. For example, many of the refs in that config value for my repos don't point to remote refs in the refs/remotes/... directory. Instead, they seem to point back to local branches, which doesn't seem to make sense. Also, in my repos that just have a single "origin" repo and are on the master branch there are no .git/refs/remotes/origin/master refs. Instead, there's just the .gif/refs/remotes/origin/HEAD files, which seem to point back at themselves.

driusan: I think it's the name *on the remote* to send to `git fetch-pack`, not the local ref name.  We probably can't do it properly until fetch-pack is implemented, but you can add a comment and rename the variable for now

sirnewton01:
@driusan I have added more changes. I think it's ready for review.

driusan:
I didn't compare it to the man page to make sure the syntax is an exact match,  but it looks good and if something comes up we can always fix it later.

PR #109, opened 2018-08-19
Add stubs with default behaviour for submodule and show-ref subcomman…

sirnewton01:
…ds making the go get tests pass

### DIFF
--- a/cmd/showref.go
+++ b/cmd/showref.go
@@ -16,6 +16,14 @@ func ShowRef(c *git.Client, args []string) error {
 		flags.PrintDefaults()
 	}
 
-	//flags.Parse(args)
+	// These flags can be moved out of these lists and below as proper flags as they are implemented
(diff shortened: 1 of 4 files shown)


cmd/showref.go
```diff
@@ -0,0 +1,21 @@
+package cmd
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/driusan/dgit/git"
+)
+
+func ShowRef(c *git.Client, args []string) error {
```
driusan: I'm not sure how I feel about having this lie about being successful just to get `go get` to not report an error, but I think I might need to implement `show-ref` to `git upload-pack` to work. Can you have this parse the options listed in the man page into a struct and I'll pick it up from there?

sirnewton01: Sure, I can do that.

For now, the rationale is that it doesn't support any refs and shows nothing. It's certainly a stretch, but it successfully shows nothing. :)

PR #110, opened 2018-08-20
Implement git mktag subcommand

driusan:
Add plumbing to create annotated tags.


git/tag.go
```diff
@@ -2,12 +2,34 @@ package git
...
+}
+
+// List tags, if tagnames is specified only list tags which match one
+// of the patterns provided.
+func TagList(c *Client, patterns []string) ([]string, error) {
+	if len(patterns) != 0 {
+		return nil, fmt.Errorf("Tag list with patterns not implemented")
+	}
+
+	files, err := ioutil.ReadDir(c.GitDir.String() + "/refs/tags")
```
sirnewton01: I like to try to use filepath.Join() in these cases, just to be sure that there isn't "//" and the delimiter is the right one for windows.

git/mktag.go
```diff
@@ -0,0 +1,145 @@
...
+	// to return a different error than the second to pass the git test suite, so we
+	// do this in two parts
+	timestampRe := regexp.MustCompile(`^ (\d+) (\+|\-)(\d+)$`)
+	timepieces := timestampRe.FindStringSubmatch(taggerPieces[3])
+	if len(timepieces) == 0 {
+		// This is all really stupid stupid, but we need to ensure that we report
+		// the same char position and error message as git to get the tests to pass.
+
+		// If the first part isn't an integer, say it's missing.
+		timestampRe2 := regexp.MustCompile(`^ (\d+)`)
```
sirnewton01: These regexes should probably be pulled out into vars so that if they don't compile you find out right away when dgit starts instead of failing only when this code path is run.

git/mktag.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Implements mktag by reading the input from r in the format
+// described in git-mktag(1).
+func Mktag(c *Client, r io.Reader) (Sha1, error) {
+	// This goes through a ridiculous amount of effort to ensure, not only that we
+	// error out in the same conditions as the official git client, but also that we
+	// error out with the exact same message (and character position), even when multiple
+	// things are wrong or the error message is misleading. The payoff is that we can run
+	// the official git test suite against it.
+	//
```
sirnewton01: Yes, this is the double edged sword for both the goal of being a drop-in replacement for official git and also to use the official git test suite. :(

git/mktag.go
sirnewton01: Ah, right. Not such a great design for ephemeral command-line tools with many subcommands in them.

PR #115, opened 2018-08-25
Add initial support for go modules

driusan:
This attempts to add basic support for go 1.11 modules by making import paths relative and adding a go.mod file.

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/driusan/dgit/git"
+	"../git"
 )
 
 func Add(c *git.Client, args []string) error {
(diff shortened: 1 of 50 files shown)


driusan:
The errors from Travis are happening because `go build` fails within $GOPATH with these changes. (I'm not entirely sure what the right way to remain go get-able is with go mod support with this repo.)

driusan:
#116 does a better job of adding go.mod with no changes to the source allows compiling outside of $GOPATH with 1.11, but I'm leaving this branch around because the bug reported in golang/go#27224 (code that compiled with 1.10 failing with 1.11) is still there.

sirnewton01:
I'm still trying to understand the new go module system. There is an article by Dave Cheney here about how to hook it into Travis for an existing project: https://dave.cheney.net/2018/07/16/using-go-modules-with-travis-ci

PR #117, opened 2018-09-02
Add support for rev-parse with a path component

### LINKED ISSUE #102: rev-parse missing treeish:file syntax
You should be able to do things like "git rev-parse foo:bar/baz" to refer to the file bar/baz in the tree foo. This isn't parsed correctly and instead looks for a commit named "foo:bar/baz" currently.

driusan:
This adds support for rev-parse of the form "git rev-parse foo:bar"
where foo is a treeish and bar is a path.

This implementation is not very efficient, but seems to match the
output of git, at least in the basic use cases.

Fixes #102

### DIFF
--- a/git/revparse.go
+++ b/git/revparse.go
@@ -264,7 +264,7 @@ func RevParse(c *Client, opt RevParseOptions, args []string) (commits []ParsedRe
 					sha = arg
 					exclude = false
 				}
-				if path := strings.Index(arg, ":"); path >= 0 {
+				if strings.Contains(arg, ":") {
 					sha, err := RevParsePath(c, &opt, arg)
 					if err != nil {
 						err2 = err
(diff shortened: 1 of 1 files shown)


git/revparse.go
```diff
@@ -193,19 +257,27 @@ func RevParse(c *Client, opt RevParseOptions, args []string) (commits []ParsedRe
...
 					sha = arg[1:]
 					exclude = true
 				} else {
 					sha = arg
 					exclude = false
 				}
-				cmt, err := RevParseCommit(c, &opt, sha)
-				if err != nil {
-					err2 = err
+				if path := strings.Index(arg, ":"); path >= 0 {
```
sirnewton01: Maybe strings.Contains() would be a bit clearer?

driusan: Sure, changed

PR #120, opened 2018-09-04
Fix read-tree --reset tests

driusan:
This fixes the read-tree tests for t1005, which test --reset.

This primarily involves adding support for `HEAD^` references (but only does the minimal to have the test pass) and removing unmerged files from the file system when resetting.


git/file.go
```diff
@@ -14,10 +14,10 @@ type File string
 
 // Determines if the file exists on the filesystem.
 func (f File) Exists() bool {
-	if _, err := os.Lstat(string(f)); os.IsNotExist(err) {
-		return false
+	if _, err := os.Lstat(string(f)); err == nil {
```
sirnewton01: How did os.IsNotExist() not work here?

git/lsfiles.go
```diff
@@ -260,13 +259,15 @@ func LsFiles(c *Client, opt LsFilesOptions, files []File) ([]*IndexEntry, error)
...
-			_, err := f.Stat()
-			// The file being deleted means it was modified
-			if os.IsNotExist(err) {
+			// If we couldn't stat it, we assume it was deleted and
+			// is therefore modified. (It could be because the file
+			// was deleted, or it could be bcause a parent directory
+			// was deleted and we couldn't stat it. The latter means
+			// that os.IsNotExist(err) can't be used to check if it
+			// really was deleted, so for now we just assume.)
+			if _, err := f.Stat(); err != nil {
```
sirnewton01: Ah, here it is. That error only works when the file itself doesn't exist, not one of the parent directories.

git/readtree.go
```diff
@@ -495,18 +502,63 @@ func readtreeSaveIndex(c *Client, opt ReadTreeOptions, i *Index) error {
 	return nil
 }
 
+// Check if the read-tree can be performed. Returns a list of files that need to be
+// removed if --reset is specified.
+func checkReadtreePrereqs(c *Client, opt ReadTreeOptions, idx *Index) ([]File, error) {
+	if opt.Update && opt.Prefix == "" && !opt.Merge && !opt.Reset {
+		return nil, fmt.Errorf("-u is meaningless without -m, --reset or --prefix")
```
sirnewton01: It seems a little weird to produce error messages in the git package that refer to flags defined in the cmd package. I'm assuming that this is here because of the official git tests checking specific error messages?

git/reset.go
```diff
@@ -96,7 +96,7 @@ unstage:
 // ResetMode implements "git reset [--soft | --mixed | --hard | --merge | --keep]" <commit>
 func ResetMode(c *Client, opts ResetOptions, cmt Commitish) error {
 	if !opts.Soft && !opts.Mixed && !opts.Hard && !opts.Merge && !opts.Keep {
-		// The default mode is mixed is none were implemented.
+		// The default mode is mixed is none were specified.
```
sirnewton01: s/is none/if none/ maybe?

git/revparse.go
```diff
@@ -182,48 +182,75 @@ func RevParseTreeish(c *Client, opt *RevParseOptions, arg string) (Treeish, erro
...
+				if newerr != nil {
+					err = newerr
+					return
+				}
+				parents, newerr := basecmt.Parents(c)
+				if newerr != nil {
+					err = newerr
+					return
+				}
+				if len(parents) != 1 {
```
sirnewton01: Ah, good point. I hadn't thought about merge commits in this context before. :)

git/readtree.go
driusan: I can't remember if there was a test requiring the wording or I just wanted it to match,  but this is just moving it from the end in checkMergeAndUpdate to a new checkPrereqs to evaluate the cheap checks before doing all the work.

git/reset.go
driusan: Fixed

PR #123, opened 2018-09-08
Optimize rev-list (and incidentally push) commands

driusan:
This does some slight optimizations to the git-rev-list command in order to try and improve the speed of "git push". It prevents reading entire objects if all you want is the size or type in various places, and prevents parsing and re-parsing trees that it will just end up throwing away as it walks the history because if it's already seen those trees (or blobs) in previous commits.

The performance is still very bad, but the benchmark of `dgit --quiet --objects HEAD ^HEAD^` goes from approximately 60 seconds to approximately 15 seconds on my machine under DragonFly. (Compared to running the command with the real git client, where it takes approximately ~0.1s.)


driusan:
Added a cache of objects to the client which brought the last test down to taking 9 seconds. Now only a few couple of magnitude away from acceptable performance..

PR #124, opened 2018-09-09
Refactor rev-list into git and cmd package

driusan:
This builds on #123 and refactors rev-list to use the cmd package for command line parsing and the git package for work. It also includes a RevListCallback that can be used to call a callback as it walks the revisions, rather than returning them directly.

The performance is similar to #123.

### DIFF
--- a/cmd/rev-list.go
+++ b/cmd/rev-list.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"flag"
 	"fmt"
+	"os"
 
 	"github.com/driusan/dgit/git"
 )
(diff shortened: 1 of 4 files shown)


cmd/rev-list.go
```diff
@@ -16,13 +17,15 @@ func RevList(c *git.Client, args []string) ([]git.Sha1, error) {
 		flags.PrintDefaults()
 	}
 
-	includeObjects := flags.Bool("objects", false, "include non-commit objects in output")
-	quiet := flags.Bool("quiet", false, "prevent printing of revisions")
```
sirnewton01: Did you want to add all of the not yet supported flags in here?

PR #130, opened 2018-09-14
Add factotum support for Plan 9 and disable password echo on other

driusan:
This improves the reading of passwords for git push by using factotum
on Plan 9, and without echoing to the terminal on other systems.

### DIFF
--- a/git/password_other.go
+++ b/git/password_other.go
@@ -1,0 +1,15 @@
+//+build !plan9
+package git
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+func getPassword(url string) (string, error) {
+	fmt.Fprintf(os.Stderr, "Password: ")
+	pwb, err := terminal.ReadPassword(0)
+	return string(pwb), err
+}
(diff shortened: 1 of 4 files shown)


git/password_plan9.go
```diff
@@ -0,0 +1,19 @@
...
+	"os/user"
+)
+
+func getPassword(url string) (string, error) {
+	user, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+
+	return libauth.Getuserpasswd(
```
sirnewton01: Thank you, I was wondering how to do this for ghfs. I'll check out mischief's library. I'm guessing that it isn't super complicated and just read/writes the correct files.

git/password_other.go
```diff
@@ -0,0 +1,16 @@
...
+package git
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+func getPassword(url string) (string, error) {
```
sirnewton01: I suppose that in the future we could add platform-specific files for other OSes and use their own password vaults.

PR #137, opened 2018-09-16
Add support for skip-worktree and sparse checkouts in read-tree

driusan:
This adds support for git read-tree with sparse checkouts, and fixes any bugs which were encountered while attempting to get the "real" git test suite to pass.


git/lsfiles.go
```diff
@@ -250,36 +253,28 @@ func LsFiles(c *Client, opt LsFilesOptions, files []File) ([]*IndexEntry, error)
...
-					entry.PathName = "R " + entry.PathName
-				}
-				fs = append(fs, entry)
+				fs = append(fs, LsFilesResult{entry, 'R'})
 				continue
 			}
 		}
 
 		if opt.Unmerged && entry.Stage() != Stage0 {
-			if opt.Status {
```
sirnewton01: 👍

official-git/run-tests.sh
```diff
@@ -9,7 +9,7 @@
 set -e
 
 # This is the tag of the release that dgit originated
-TAG=v2.8.0
```
sirnewton01: If we are testing dgit against the official git 2.10 tests then can we state that we are compatible/aligned with that version?

driusan: Sure, at some point we should see how high we can push it now that we have a reasonable number of tests running, 2.10 was just the minimum I had to push it in order to not have the test suite fail because it was insisting on having a grammar error

driusan: (But "compatible" is a strong word to use...)

PR #138, opened 2018-09-17
Improved factotum, vendor libauth

driusan:
Use factotum to get both the username and the password on Plan 9, and add to the vendor directory.


driusan:
(Merged without giving time for anyone to look at it because it also fixes the the build with upstream changes to libauth..)

PR #159, opened 2018-10-08
Replace io.Copy with TeeReader when indexing a packfile

driusan:
dgit was doing an io.Copy to download the data from a network connection to a temporary file, and then indexing based on that temp file so that it could use an the file as a io.ReadSeeker rather than an io.Reader. This replaces the copy with an io.TeeReader in index pack which builds the temp file as it's being read. This means that indexing of packs can begin immediately when doing a clone or fetch, rather than having to wait until all the data is downloaded before starting. Theoretically, this should improve performance on large repositories (if you have enough memory to hold the resolved deltas) since the index-pack gets a head start.


git/httpconn.go
```diff
@@ -337,6 +339,7 @@ func (s *smartHTTPConn) sendRequest(expectedmime string) error {
 	}
 
 	s.lastresp = resp.Body
+	fmt.Printf("Setting connection")
```
sirnewton01: Was this meant to be log.Printf()?

driusan: No, it's meant to be deleted, I just missed it..

PR #177, opened 2018-10-24
Added git update-index --index-info missing variant. Fixes #43

### LINKED ISSUE #43: git update-index --index-info missing 1 variant
Of the 3 variants of `--index-info` listed in `git-update-index(1)` only the one used by the t0000-basic.sh test suite is implemented.

It should be fairly to add support for the other two variants by basing them on the one which is implemented in UpdateIndexFromReader and adding new test cases.

guidopola:


git/updateindex.go
```diff
@@ -142,7 +142,22 @@ func UpdateIndexFromReader(c *Client, opts UpdateIndexOptions, r io.Reader) (*In
 		switch len(spaces) {
 		case 2:
 			// mode SP sha1 TAB path
-			return nil, fmt.Errorf("update-index --index-info variant 1 not implemented")
+			if spaces[0] == "0" {
+				// Mode "0" means remove index
```
driusan: From the example I see at https://www.git-scm.com/docs/git-update-index, it reads to me like this check both that the sha1 is the zero sha1 and the mode is 0, but the wording is a little ambiguous so I'm not sure

guidopola: Well it says it doesn't matter as long as it is well formatted, i will add a check to validate the sha1 input
> The first line of the input feeds 0 as the mode to remove the path; the SHA-1 does not matter as long as it is well formatted

PR #180, opened 2018-11-03
Partial implementation of git-archive

guidopola:
While i do understand that Plumbing commands are a priority I partially implemented the git-archive command to better understand the codebase.

### **What's missing?**
--verbose
--worktree-attributes
--remote
--exec
\<path> optional arg

**Usage of config options**
tar.umask
tar.\<format>.command
tar.\<format>.remote
symlinks support
git submodule?

**All of the examples in the git docs works except of this:**
`git archive TAG fails if the ref file [refs/tags/TAG] doesn't exist.`

If you clone a repo the ref file doesn't exist, looks like there is something not implemented yet because dgit tag also shows nothing.

`git archive --format=zip --prefix=git-docs/ HEAD:Documentation/ > git-1.4.0-docs.zip`

`git archive --format=tar --prefix=git-1.4.0/ v1.4.0^{tree} | gzip >git-1.4.0.tar.gz `
(_are the ^{tree} modifier not implemented?_)

`git config tar.tar.xz.command "xz -c" `

### DIFF
--- a/cmd/archive.go
+++ b/cmd/archive.go
@@ -4,43 +4,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/driusan/dgit/git"
 )
-
-func argsParseCompressionLevel(opts *git.ArchiveOptions, args []string) []string {
-	for i, arg := range args {
-		if len(arg) < 2 || arg[0] != '-' {
-			continue
-		}
-		startIndex := 1
-		if arg[1] == '-' {
-			startIndex++
-			if len(arg) == 2 {
-				continue
-			}
-		}
-
-		// slice the dash
-		value := arg[startIndex:]
-
-		// Try to convert to integer
-		if v, err := strconv.Atoi(value); err != nil {
-			// The value is NaN
-			continue
-		} else {
-			// The compression value must be between 0..9
-			if v >= 0 && v <= 9 {
-				opts.CompressionLevel = v
-			}
-			// remove this value from the arg list
-			return append(args[:i], args[i+1:]...)
-		}
-	}
-	return args
-}
 
 func Archive(c *git.Client, args []string) error {
 	flags := flag.NewFlagSet("archive", flag.ExitOnError)
@@ -52,6 +19,9 @@ func Archive(c *git.Client, args []string) error {
 	}
 
 	opts := git.ArchiveOptions{}
+
+	// default compression level in the deflate package.
+	opts.CompressionLevel = -1
 
 	//flags.BoolVar(&opts.Verbose, "verbose", false, "Report archived files on stderr")
 	//flags.BoolVar(&opts.Verbose, "v", false, "Alias for --verbose")
@@ -73,14 +43,18 @@ func Archive(c *git.Client, args []string) error {
 
 	flags.StringVar(&opts.ArchiveFormat, "format", "", "Archive format")
 
-	flags.IntVar(&opts.CompressionLevel, "0", 0, "Store only")
-	flags.IntVar(&opts.CompressionLevel, "9", 9, "Highest compression level")
-
-	// if the args is empty, print usage.
-	if len(args) == 0 {
-		flags.Usage()
-		os.Exit(2)
-	}
+	// FIXME: Find a better way to do this.
+	var cl [10]bool
+	flags.BoolVar(&cl[0], "0", false, "No compression")
+	flags.BoolVar(&cl[1], "1", false, "Compress faster")
+	flags.BoolVar(&cl[2], "2", false, "Compression level")
+	flags.BoolVar(&cl[3], "3", false, "Compression level")
+	flags.BoolVar(&cl[4], "4", false, "Compression level")
+	flags.BoolVar(&cl[5], "5", false, "Compression level")
+	flags.BoolVar(&cl[6], "6", false, "Compression level")
+	flags.BoolVar(&cl[7], "7", false, "Compression level")
+	flags.BoolVar(&cl[8], "8", false, "Compression level")
+	flags.BoolVar(&cl[9], "9", false, "Highest compression level")
 
 	flags.Parse(args)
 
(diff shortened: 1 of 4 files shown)


driusan:
Wow, thanks!

When the documentation plumbing commands are a priority, I just mean that they're my priority because they make implementing other commands easier and more consistent. For archive I don't think there's much plumbing that would be re-useable in the first place, anyways.

cmd/archive.go
```diff
@@ -0,0 +1,133 @@
...
+	flags.BoolVar(&opts.List, "list", false, "List supported archive formats")
+	flags.BoolVar(&opts.List, "l", false, "Alias for --list")
+
+	flags.Var(newNotimplStringValue(), "remote", "Not implemented")
+	flags.Var(newNotimplStringValue(), "exec", "Not implemented")
+
+	flags.StringVar(&opts.ArchiveFormat, "format", "", "Archive format")
+
+	flags.IntVar(&opts.CompressionLevel, "0", 0, "Store only")
+	flags.IntVar(&opts.CompressionLevel, "9", 9, "Highest compression level")
```
driusan: You're missing 1-8 here (the man page only lists 0 and 9 explicitly, but the description for 9 says "You can specify any number from 1 to 9 to adjust compression speed and ratio.")

git/archive.go
```diff
@@ -0,0 +1,274 @@
...
+	ArchiveFormat      string
+	BasePrefix         string
+	OutputFile         string
+	CompressionLevel   int
+}
+
+type ArchiveFormat int
+
+const (
+	archiveUnknow = ArchiveFormat(iota)
```
driusan: archiveUnknown

git/archive.go
```diff
@@ -0,0 +1,274 @@
...
+	// otherwise we must use the commit time.
+	if sha.Type(c) != "tree" {
+		if t, err := CommitID(sha).GetDate(c); err == nil {
+			mtime = t
+		}
+	}
+
+	var tw *tar.Writer
+
+	//
```
driusan: forgot to delete this or write something?

git/archive.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		if obj, ok := o.(GitBlobObject); ok {
+			hdr := &tar.Header{}
+			hdr.Name = opts.BasePrefix + e.PathName.String()
+			hdr.Size = int64(obj.GetSize())
+			hdr.ModTime = mtime
+			hdr.Uid = 0
+			hdr.Gid = 0
+			hdr.Uname = "root"
+			hdr.Gname = "root"
```
driusan: Are these permanent or just a stub? If a stub, there should probably be a comment explaining what real git does.

cmd/archive.go
guidopola: They are parsed in the [argsParseCompressionLevel](https://github.com/driusan/dgit/blob/4ccf41a07eedc0902e948cf6f607489e6849cdb6/cmd/archive.go#L13) function outside of the go flags to avoid doing command line usage a lot verbose, for example official git client only shows options 0 and 9.
though i can change it to use the FlagSet if you want

git/archive.go
guidopola: Yeah.. This is because I'm very used to code in C++ (only one return statement). 
I'll update the code to be more golang friendly :+1:

git/archive.go
guidopola: It's declared here because to use gzip compression the tar.Writer output must be set to a gzip.Writer.
See [this code](https://github.com/driusan/dgit/blob/4ccf41a07eedc0902e948cf6f607489e6849cdb6/git/archive.go#L82)

git/archive.go
guidopola: The official client set the file owner and group name to root but this is probably not needed.

git/archive.go
guidopola: This can be changed to the code below, but in my opinion the defer thing is a mess.
```
var fileOutput io.Writer = os.Stdout
...
// If gzip compression is enabled
if tgz {
	gw := gzip.NewWriter(fileOutput)
	fileOutput = gw
	defer gw.Close()
}

// Create the tar writer
tw := tar.NewWriter(fileOutput)
defer tw.Close()
```

driusan: No, I just mean the empty "//" delimiter, not the code

guidopola: I just forgot to remove the empty comment :-)

git/archive.go
```diff
@@ -0,0 +1,260 @@
...
+	for key := range supportedArchiveFormats {
+		formats = append(formats, key)
+	}
+	return formats
+}
+
+func Archive(c *Client, opts ArchiveOptions, arg string) error {
+	format := archiveTar
+
+	//
```
driusan: another phantom comment delimiter

git/archive.go
```diff
@@ -0,0 +1,260 @@
...
+// Return the list of supported archive file format
+func ArchiveFormatList() []string {
+	formats := make([]string, 0, len(supportedArchiveFormats))
+	for key := range supportedArchiveFormats {
+		formats = append(formats, key)
+	}
+	return formats
+}
+
+func Archive(c *Client, opts ArchiveOptions, arg string) error {
```
driusan: If you make arg a Treeish instead of string and move the conversion into package string, you could simplify this code. (You could also implement paths by just adding a []File argument here that you pass along the LsTree..)

git/archive.go
```diff
@@ -0,0 +1,260 @@
...
+		} else {
+			treeish = c
+		}
+	}
+
+	if s, ok := treeish.(CommitID); ok {
+		sha1 = Sha1(s)
+	} else {
+		return fmt.Errorf("Can't convert treeish to commit id")
+	}
```
driusan: I think you're going to end up with a lot of cases/exceptions here. You probably just want to check if the type can be converted to a Commitish to get the CommitID with something like:

```
commitish, ok := treeish.(Commitish)
if ok {
      cid, err := commitish.CommitID(c)
      if err != nil {
            return err
      }
      sha1 = Sha1(cid)
}
```

but if I understand what you're doing correctly you're just using the sha1 to get the mtime, so you might also want to lift the mtime conversion all the way up to the exported Archive function and pass it down instead of passing the Sha1 down and duplicating it for zip/tar.

git/archive.go
```diff
@@ -0,0 +1,260 @@
...
+
+	lstree, err := LsTree(c, LsTreeOptions{Recurse: true}, treeish, nil)
```
driusan: FYI I think the implement paths all it would take is passing a []File instead of a nil here

git/archive.go
guidopola: It's not just to get the mtime but to set the sha1 value as a comment if the output is zip or as a pax object if it's tar

driusan: Ah, okay, in that case you still need to pass it, but you'll still make your life easier by lifting it up a level and checking if it implements the Conmitish interface rather than trying to exhaustively check every possible Treeish implementer type

git/archive.go
```diff
@@ -0,0 +1,260 @@
...
+	"time"
+)
+
+type ArchiveOptions struct {
+	Verbose            bool
+	List               bool
+	WorktreeAttributes bool
+	ArchiveFormat      string
+	BasePrefix         string
+	OutputFile         string
```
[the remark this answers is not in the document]
guidopola: done!

git/archive.go
guidopola: removed :+1:

git/archive.go
guidopola: Its updated now :-)

driusan: You still have the tree.(Ref) above. Ref implements Commitish, so it's redundant

guidopola: Done!

driusan: Thanks.. it all looks much cleaner now, I think.

PR #187, opened 2018-12-05
Add remote show and remote get-url subcommands

driusan:
This adds the `git remote show -n` subcommand, as well as `remote get-url`.

### DIFF
--- a/cmd/remote.go
+++ b/cmd/remote.go
@@ -35,6 +35,25 @@ func Remote(c *git.Client, args []string) error {
 		}
 		aopts := git.RemoteAddOptions{opts}
 		return git.RemoteAdd(c, aopts, args[1], args[2])
+	case "get-url":
+		uflags := newFlagSet("remote-get-url")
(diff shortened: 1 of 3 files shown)


cmd/remote.go
```diff
@@ -35,6 +35,25 @@ func Remote(c *git.Client, args []string) error {
 		}
 		aopts := git.RemoteAddOptions{opts}
 		return git.RemoteAdd(c, aopts, args[1], args[2])
+	case "get-url":
+		uflags := newFlagSet("remote-get-url")
+		urlopts := git.RemoteGetURLOptions{RemoteOptions: opts}
+		uflags.BoolVar(&urlopts.Push, "push", false, "Print push URLs, not fetch URLs")
+		uflags.BoolVar(&urlopts.All, "all", false, "Print all URLs, not just the first")
```
sirnewton01: The 'all' option doesn't seem to be used here. Maybe the description should indicate that it is not yet supported?

driusan: sure, done..

PR #189, opened 2019-02-21
Skip the vendor directories with the formatter check

sirnewton01:
It seems that there was a change to the go formatter that is tripping up on one fo the vendored libraries. This focuses the formatter checks on the project's own source code.


driusan:
Thanks.. I think the underlying root cause is that gofmt formatting changed between versions, but if it's only the vendor directory affected for now we can draw it out a little longer while keeping the check.

PR #191, opened 2019-02-25
Add NetBSD support for ctime and inode syscalls

jboverfelt:
NetBSD supports ctime but the struct field is slightly different - this adds a new file_ctime_netbsd.go and updates the build tags accordingly. Also, NetBSD was not included in file_inode_unix.go, but it supports the inode field without issue.

I've build and run this version on a 64-bit version of NetBSD 8.0 and it worked. You can of course test that it compiles by setting GOOS=netbsd as you normally would for cross-compilation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ go_import_path: github.com/driusan/dgit
 
 script:
     - GOOS=darwin go build
+    - GOOS=netbsd go build
     - GOOS=plan9 go get ./...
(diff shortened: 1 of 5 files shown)


driusan:
Thanks! Could you add a GOOS=netbsd build to the .travis.yml to ensure we don't have any regressions later on? Go builds are cheap enough that it shouldn't affect the travis time much.

jboverfelt:
Will do!

PR #195, opened 2019-02-27
Add hierarchical branch name support and extra check against HEAD branch

sirnewton01:
This was an effort to get the t3200-branch.sh official tests working. Most are still failing, but this made a few more work. I don't think it's worth adding the suite and the exclusions at this time. There's still 86/112 failing. Before going further down the rabbit hole the cmd/branch.go should probably be refactored so that the logic in the command it pulled down into the git package using the standard pattern.


git/client.go
```diff
@@ -266,6 +266,15 @@ func (c *Client) CreateBranch(name string, commit Commitish) error {
 		return err
 	}
 
+	if name == "HEAD" {
+		return fmt.Errorf("fatal: 'HEAD' is not a valid branch name.")
+	}
+
+	// Create the file so that any parent directories are created
```
driusan: I'm not sure that I understand this sentence. I think it's trying to imply that `Create` creates parents directories while the `WriteFile` below doesn't, but I'm not even sure if that's true (`Create` is a thin wrapper around `os.Create`, and `os.Create`'s doesn't say anything about what happens if the parent directory(s) aren't present)

sirnewton01: This Create() will do an os.MkdirAll() creating any necessary parent directories while WriteFile doesn't. Should I fix WriteFile so that it does?

driusan: nevermind, I was looking at GitDir.Create (which doesn't do a MkdirAll), not File.Create (which does), so the code makes sense.

I still think the wording of the comment can be cleaned up a bit. How about something like:
```suggestion
	// Create the file using File.Create first to ensure any parent directories are created
	// before writing the file contents.
```

.travis.yml
```diff
@@ -23,4 +23,4 @@ script:
...
     # t3000.7 requires git pack-refs
-    - GIT_SKIP_TESTS='t0008.30[5-9] t0008.310 t0008.321 t0008.323 t0008.37[0-9] t0008.38[0-7] t0008.39[1-2] t1004.1[6-7] t1014.26 t3000.7 t2018.[6-7] t2018.[9] t2018.1[5-8]' ./official-git/run-tests.sh t0000-basic.sh t0004-unwritable.sh t0007-git-var.sh t0008-ignores.sh t0010-racy-git.sh t0062-revision-walking.sh t0070-fundamental.sh t0081-line-buffer.sh t1000-read-tree-m-3way.sh t1001-read-tree-m-2way.sh t1002-read-tree-m-u-2way.sh t1003-read-tree-prefix.sh t1004-read-tree-m-u-wf.sh t1005-read-tree-reset.sh t1008-read-tree-overlay.sh t1009-read-tree-new-index.sh t1011-read-tree-sparse-checkout.sh t1012-read-tree-df.sh t1014-read-tree-confusing.sh t1100-commit-tree-options.sh t1304-default-acl.sh t1403-show-ref.sh t2000-checkout-cache-clash.sh t2001-checkout-cache-clash.sh t2002-checkout-cache-u.sh t2006-checkout-index-basic.sh t2009-checkout-statinfo.sh t2014-switch.sh t2018-checkout-branch.sh t2100-update-cache-badpath.sh t3000-ls-files-others.sh t3002-ls-files-dashpath.sh t3006-ls-files-long.sh t3020-ls-files-error-unmatch.sh t3800-mktag.sh t4113-apply-ending.sh t4123-apply-shrink.sh t7062-wtstatus-ignorecase.sh t7511-status-index.sh
```
sirnewton01: Due to recent efforts we can stop skipping 6 ignore tests.

driusan: I wonder if we can split this over multiple lines with `\` to get better diffs here, it's always hard to tell what's added and what's removed..

driusan: is this t0008.305 through t0008.310 being added?

sirnewton01: Yes, that's right. I'm not a yaml expert. Do multi-line strings work that way here?

sirnewton01: These might work: https://yaml-multiline.info

I can give it a try sometime.

driusan: If you want to revert so that it passes we can figure the spacing out later

sirnewton01: I pushed the list of ignores and the list of test suites into the shell script where it can be run in development environment and in travis environment more easily.

PR #196, opened 2019-02-27
Unglue flags that can be glued to satisfy golang flags

sirnewton01:
Pre-emptively unglue flags that are allowed to be glued together in the official git client to satisfy the Go flag package. Remove unneeded patch on the official test suite.


cmd/commit.go
```diff
@@ -68,7 +68,17 @@ func Commit(c *git.Client, args []string) (string, error) {
...
+	for _, a := range args {
+		// Unglue any glued -m arguments
+		if strings.HasPrefix(a, "-m") && a != "-m" {
+			adjustedArgs = append(adjustedArgs, "-m", a[2:])
+			continue
+		}
+		adjustedArgs = append(adjustedArgs, a)
+	}
+
+	flags.Parse(adjustedArgs)
```
driusan: Nice, I like this approach.

PR #198, opened 2019-02-28
Fix the flags with optional values in the status command

### LINKED ISSUE #27: Error when providing --porcelain to the status subcommand without a value
When invoking ```dgit status --porcelain somedir``` fails with an error where it cannot parse somedir as an integer. The problem is the the porcelain switch is parsed as an IntVar, but in this case it interprets somedir as the value to the porcelain switch. In the standard git the porcelain switch has an optional value and that value _must_ be provided after an equal sign resolving the ambiguity. No value is interpreted to be the value 1.

sirnewton01:
The status command has --porcelain and two other flags that can be given on the command-line with an optional value. The git manual provides default values in case the value is not provided. Dgit currently requires an explicit value because of limitations in the way that the flag package works in Go. This PR works around the problem by pre-processing the set of subcommand arguments and inserting the default values when the bare flag is provided on the command line.

This should fix #27


sirnewton01:
@driusan is it ok if I merge this? Any concerns with it?

sirnewton01:
I'm going to go ahead and merge. I think it is low risk and small. If needed, we can back it out later.

driusan:
The code looks fine, but I'm going to back it out and re-commit it because it looks like you did a merge commit instead of a squash and merge on the GitHub front end

driusan:
Okay, I think it's fine now but you might get errors saying something's fishy since I rewrote history.

sirnewton01:
Sorry @driusan I'll be more careful with the big green button next time. Any chance that the project settings can default to squash instead of merge?

driusan:
Don't worry about it,  I just like to try and keep a clean history and it's easiest to squash.. I don't think you can change the default on a repo, I think it remembers the last thing you did on a per user basis for the default,  but I disabled merge commits with the big green button for now

PR #199, opened 2019-02-28
Fix the "init --bare", the detection of bare repos and a bug with ls-remote

sirnewton01:
Initializing a bare repository didn't set the configuration properly. Also, the detection of a bare repo didn't work and falsely reporting that it wasn't a git repo. Finally, ls-remote was not outputting anything with the "-q" option, which is only supposed to squelch the "From <url>" line that gets echoed to stderr.


cmd/lsremote.go
```diff
@@ -57,10 +57,8 @@ func LsRemote(c *git.Client, args []string) error {
 	if opts.ExitCode && len(refs) == 0 {
 		os.Exit(2)
 	}
-	if !opts.Quiet {
-		for _, ref := range refs {
-			fmt.Println(ref.TabString())
-		}
+	for _, ref := range refs {
```
driusan: Are there any test cases for this (either from the official tests or `go test`)? I suspect we'll eventually forget and have a regression if it's just go modules catching it

sirnewton01: There is an ls-remote official test suite, but it hangs in the middle. It gets stuck on a local clone trying to do some kind of RPC with git-upload-pack on a pipe.

If I can get the go modules stuff working then I would add it to the go-get tests that we have and it should cover it.

driusan: Fair enough

PR #203, opened 2019-03-01
Add global -c parameter to support session-scoped configuration values

sirnewton01:
Fixes #203


driusan:
... but not to Travis. You missed a reference in one of the tests

PR #204, opened 2019-03-01
Add max-count options to git log to limit the number of commits shown

sirnewton01:


cmd/log.go
```diff
@@ -85,8 +106,19 @@ func Log(c *git.Client, args []string) error {
 	if err != nil {
 		return err
 	}
-	return git.RevListCallback(c, git.RevListOptions{Quiet: true}, []git.Commitish{commit}, nil, func(s git.Sha1) error {
+
+	var numCommits = 0
+
+	err = git.RevListCallback(c, git.RevListOptions{Quiet: true}, []git.Commitish{commit}, nil, func(s git.Sha1) error {
```
driusan: I think the max-count is supposed to be a parameter to rev-list that log just passes down.. can we move this into RevListCallback?

sirnewton01: Thanks, yes it looks like the pretty printing should be pushed down there too. It looks like the printing of the output will be the only thing that log command will do in the future.

git/revlist.go
```diff
@@ -8,8 +8,11 @@ import (
 // List of command line options that may be passed to RevList
 type RevListOptions struct {
 	Quiet, Objects bool
+	MaxNumCommits  *int
```
driusan: I think we should stick to the same terminology as the command line ("MaxCount"), because the terminology MaxNumCommits isn't accurate when Objects is true.

Also, I think it it should be either `*uint` or `int` (with a convention of a negative number means "all".. I think I prefer the latter.)

sirnewton01: Sure, I'll change the name. The git docs here talk about number commits specifically so I thought that it was always about commits: https://git-scm.com/docs/git-rev-list

I was going to use -1 as the convention but then I was thinking that the pointer makes better use of the zero value as the default (show me everything). If I switch it to *uint it will also prevent any really nonsensical negative numbers. I'll switch it to *uint unless you have reservations.

driusan: The only reservation I have against `*uint` is that then you've always gotta check for nil before dereferencing, while `int` is only one check, either make sense to me.

PR #206, opened 2019-03-01
Add format option to log subcommand, add some common formats, add ref name list for commit

sirnewton01:
This is a start for the formatting of commits in git log that can be applied elsewhere in the future. Also, the formats supported can be expanded over time. Can you have a look and see if I'm on the right track? Also, this should get us closer to supporting "go get" with modules.


git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
 }
+
+// Internal pre-computed record of which symbolic references
+//  are pointing to a particular ref name (without the refs/ prefix)
+var symRefsForRefName map[string]string
+
+// Internal pre-computed record of which ref names are pointing to
+//  to a particular commit. Values are in the form of "HEAD -> master, branch2,
+//  tag: tag1"
+var refNamesForCommit map[string]string
```
driusan: Are you sure that these are required, or are you just assuming they are? I know I had to rip out some caches (mostly with delta calculation) a while ago because while they helped with smaller repos, they ate up all the memory and crashed on bigger repos.

If they are required, I think that:
1. They should use better types than strings
2. They should be on the Client object instead of globals for thread safety.

git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
+
+	branches, err := c.GetBranches()
+	if err != nil {
+		return err
+	}
+
+	for _, branch := range branches {
+		branchName := string(branch)
+
+		commitId, err := RefSpec(branchName).Value(c)
```
driusan: We should be more careful about "RefSpec" vs "Ref" (although there's still a lot of places in older code that use the word "RefSpec" instead of "Ref")..a Ref is a single reference in a repo. A RefSpec is (or at least should be) a spec that can include wildcards, but is often used incorrectly in this repo from before I understood the difference.

git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
+var symRefsForRefName map[string]string
+
+// Internal pre-computed record of which ref names are pointing to
+//  to a particular commit. Values are in the form of "HEAD -> master, branch2,
+//  tag: tag1"
+var refNamesForCommit map[string]string
+
+// We have been asked to pre-compute the ref names for important
+//  commits as a way to improve performance.
+func buildRefNames(c *Client) error {
```
driusan: Can most of this function be replaced with a call to ShowRef? If so, it might be a good idea to use the git plumbing that's already there.

git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
+	}
+
+	return nil
+}
+
+// Since libgit is somewhat out of our control and we can't implement
+// a fmt.Stringer interface there, we use this helper.
+
+// Print commit using "medium" format
+func printCommitMedium(c *Client, cmt CommitID, format string) error {
```
driusan: Are these required? Can they be replaced with pre-defined formatting strings for printCommitFormat?

The only hiccup I'd see is merge parents, but (if git doesn't already have one) we can temporarily add a "print all merge commit IDs" verb until the looping stuff is there.

git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
+	return nil
+}
+
+func printCommitOneLine(c *Client, cmt CommitID, format string) error {
+	msg, err := cmt.GetCommitMessage(c)
+	if err != nil {
+		return err
+	}
+	// TODO figure out if merge parents should show up in this mode or not
+	title := strings.Split(strings.TrimSpace(msg.String()), "\n")[0]
```
driusan: This needs a bounds check for empty messages, otherwise it'll panic.

cmd/log.go
```diff
@@ -111,8 +75,13 @@ func Log(c *git.Client, args []string) error {
 		opts.MaxCount = &mc
 	}
 
+	printCommit, err := git.GetCommitPrinter(c, format)
```
driusan: maybe instead we should have a `Commit.Format(c, format) (string, error)` function and use that inside the unexported printCommit? `GetCommitPrinter` feels like an internal implementation thing that shouldn't be exported, while `Format` feels like it belongs in on the object.

git/revlist.go
sirnewton01: Agreed that this could be premature optimization without any benchmarks. I felt that reading the ref and head files over and over for each commit in history might be bad performance, but perhaps I am underestimating OS caches.

Fair enough, I can remove these for now until we notice a performance problem. If they are needed, the caches could be managed and typed better.

git/revlist.go
sirnewton01: Sure, I'll have a look. I wasn't sure if we had something that calculated the ref names of a commit in dgit already.

git/revlist.go
sirnewton01: Yes, I think I now see the difference. A Ref is a pointer file in the .git/refs directory. A RefSpec is something that may or may not match Ref. I think that RefSpec can include both source and destinations as I can see from your PR and after reading more in the git manuals.

git/revlist.go
sirnewton01: Yes, I'm wondering if these can be replaced by formats. Perhaps it could all be put into a Commit type as a Format() method.

git/revlist.go
sirnewton01: 👍

git/revlist.go
driusan: Yeah, there's a lot of places in the code that need that cleaned up, because I didn't really understand that at first either. At some point it might be worth just ripping out everything other than "Src" and "Dst" from the Refspec type and then fixing all the compile errors will force us to be more careful.

(Also, refs can be packed with newer versions of git, so they don't necessarily live under .git/refs, but conceptually.. yeah.)

git/revlist.go
driusan: You're probably right, and caching every reference is cheaper than caching every commit (which is where we had problems before).. but I still think it should be on Client so that multiple client's can use the library concurrently without bumping into each other's cache.

git/revlist.go
```diff
@@ -117,3 +120,207 @@ func revListCallback(c *Client, opt RevListOptions, commits []CommitID, excludeL
...
+	output := format[7:]
+
+	// Full commit hash
+	if strings.Contains(output, "%H") {
+		output = strings.Replace(output, "%H", cmt.String(), -1)
+	}
+
+	// Committer date as unix timestamp
+	if strings.Contains(output, "%ct") {
+		date, err := cmt.GetDate(c)
```
driusan: This is the wrong date.. it's returning the author timestamp, not the committer

cmd/log.go
sirnewton01: It's not possible to convert all of the pre-built formats, such as medium, into pretty print formats, so we can't get away with just a single Commit.Format() method. I'm thinking that if we can have a handful of methods that this will be ok: Commit.[Medium|Full|Fullest]Format(), Commit.Format()

driusan: Okay, that seems reasonable.. they don't need to be exported, though.  Format can just delegate to the other functions if passed one of the special keywords. 

Out of curiosity, what's the reason they can't be?

sirnewton01: Medium, for example has the "Merges:" part that can't be described using the pretty format tags. That isn't necessarily an issue as it can be injected by the command. The spec defines that the Merges: section should come immediately before "Author:" This is a hack, but doable.

The problem comes with the indenting of the commit messages. There's no apparent way with a format string to specify an indent for each line of the output that is being interpolated with the tag.

cmd/fetchpack.go
```diff
@@ -27,7 +27,7 @@ func FetchPack(c *git.Client, args []string) error {
 	flags.BoolVar(&opts.Verbose, "verbose", false, "Be more verbose")
 	flags.Parse(args)
 	args = flags.Args()
-	opts.Depth = int32(*depth)
+	opts.Depth = int(*depth)
```
driusan: I don't think we do anything with this, maybe we should document that's it's only parsed and not implemented?

git/fetchpack.go
```diff
@@ -26,7 +26,7 @@ type FetchPackOptions struct {
 	Thin                           bool
 	IncludeTag                     bool
 	UploadPack                     string
-	Depth                          int32
+	Depth                          int
```
driusan: Why is this changed from int32 to int? The officially documentation implies that it's a 32 bit integer:

```
Limit fetching to ancestor-chains not longer than n. git-upload-pack treats the special depth 2147483647 as infinite even if there is an ancestor-chain that long
```

git/sha1.go
```diff
@@ -430,6 +498,109 @@ func (c CommitID) GetAllObjectsExcept(cl *Client, excludeList map[Sha1]struct{})
...
+		output = strings.Replace(output, "%ct", strconv.FormatInt(date.Unix(), 10), -1)
+	}
+
+	// Author date as unix timestamp
+	if strings.Contains(output, "%at") {
+		date, err := c.GetDate(cl)
+		if err != nil {
+			return "", err
+		}
+		output = strings.Replace(output, "%ct", strconv.FormatInt(date.Unix(), 10), -1)
```
driusan:
```suggestion
		output = strings.Replace(output, "%at", strconv.FormatInt(date.Unix(), 10), -1)
```

typo, I think

git/showref.go
```diff
@@ -54,6 +54,18 @@ func (r Ref) TabString() string {
 	return fmt.Sprintf("%v\t%v", r.Value, r.Name)
 }
 
+// Gets the "ref name" of the ref suitable for showing in places
+//  like git log and git rev-list.
+func (r Ref) RefName() string {
```
driusan: I would expect "RefName" method to return  `Refname(r.Name)`

Then you can add a ShortName` (or whatever git documentation refers to the short name as..) method to Refname and then use that.

git/fetchpack.go
sirnewton01: I did this so that we could pass a pointer to the option to flags.IntVar(). Otherwise, we need a temp variable and then a conversion. I'll put this back the the way it was.

git/showref.go
sirnewton01: The git documentation calls these short form names ref names. See documentation of %D here: https://git-scm.com/docs/git-rev-list

driusan:
I think this is probably good enough for our purposes, if needed we can refactor later

PR #209, opened 2019-03-03
Add "git fetch remote refspec" support

driusan:
Add initial support for passing a refspec on the command line
to "git fetch".

This adds support for one of the formats of "git fetch" which
is invoked by Go when working outside of $GOPATH in go modules
mode.

### DIFF
--- a/git/refspec.go
+++ b/git/refspec.go
@@ -19,6 +19,45 @@ func (r RefSpec) String() string {
 	return strings.TrimSpace(strings.TrimSuffix(string(r), "\000"))
 }
 
+// Src represents the ref name of the ref at the remote location for a refspec.
+// for instance, in the refspec "refs/heads/foo:refs/remotes/origin/foo",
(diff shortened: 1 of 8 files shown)


git/refspec.go
```diff
@@ -19,6 +19,45 @@ func (r RefSpec) String() string {
...
+// for instance, in the refspec "refs/heads/foo:refs/remotes/origin/foo",
+// src refers to refs/heads/foo, the name of the remote reference, while dst
+// refers to refs/remotes/origin/master, the location that we store our cache
+// of what that remote reference is.
+func (r RefSpec) Src() Refname {
+	if len(r) < 1 {
+		return ""
+	}
+
+	if r[0] == '+' {
```
sirnewton01: I suppose we are slamming (forcing) updates to dest in all cases for now, so this just strips it off and ignores it.

driusan: even if we weren't, I think Src() shouldn't include it, there should be a different `IsForced()` method when we get there.

PR #210, opened 2019-03-03
Add ability to refer to commits by a sha1 prefix

driusan:
This adds the ability to refer to commits on the command line
by any unique prefix of the commit id which is at least 3
characters long.

### DIFF
--- a/git/revparse.go
+++ b/git/revparse.go
@@ -2,7 +2,9 @@ package git
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


git/revparse.go
```diff
@@ -270,7 +272,67 @@ func RevParseCommitish(c *Client, opt *RevParseOptions, arg string) (cmt Commiti
...
-	return nil, fmt.Errorf("Could not parse %v", arg)
+
+	// Try seeing if it's an abbreviation of a commit as a last
+	// resort. We require a length of at least 3, so that we only
+	// need to search one directory of the objects directory.
+	if len(cmtbase) > 2 && len(cmtbase) < 40 {
+		dir := cmtbase[:2]
+		var candidates []CommitID
+
+		files, err := ioutil.ReadDir(c.GitDir.File(File("/objects/" + dir)).String())
```
sirnewton01: I wonder if we should be trying to use filepath.Join() more often when composing paths.

driusan: sure, I'll fix it tonight when I'm home from work.

driusan: Done.. as a bonus, it also looks cleaner to read

PR #211, opened 2019-03-05
Make GitDir.Create() create parent directory

driusan:
This updates c.GitDir.Create() to create the parent directory of the
file if it doesn't exist, similarly to how File.Create does.

This serves 2 purposes:

1.  Make the 2 create functions symmetrical
2.  Causes UpdateRef (which uses GitDir.Create()) to create 
    the parent directory

Without this change, if you do "git remote add something; git fetch
something" it fetches all the objects from remote "something", and
then crashes with an error about the parent directory 
`refs/remotes/something` not existing before creating the references.

### DIFF
--- a/git/client.go
+++ b/git/client.go
@@ -215,7 +215,14 @@ func (gd GitDir) Open(f File) (*os.File, error) {
 // Creates a file relative to GitDir. There should not be
 // a leading slash.
 func (gd GitDir) Create(f File) (*os.File, error) {
-	return os.Create(gd.String() + "/" + f.String())
+	fpath := filepath.Join(gd.String(), f.String())
+	dir := File(filepath.Dir(fpath))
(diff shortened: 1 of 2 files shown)


sirnewton01:
I'm assuming that you're trying the new "go get" with these change. Can you check that "git init --bare" is still working with this change? I haven't added a test yet.

driusan:
You were right,  I shouldn't have included the change to main.go

sirnewton01:
> You were right, I shouldn't have included the change to main.go

Thank you for trying it out. We definitely have some gaps in our testing. I'm hoping that once we get "go get" working regressions in some of these areas will be caught by that test. As the puzzle pieces come together some of these might be covered by the official tests too. There's just so many tests that rely on a fully functioning git to work.

PR #212, opened 2019-03-05
Fix ls-tree (and dependent commands) in bare mode

driusan:
LsTree tries to only display things that are relative
to the current working directory. In a bare repo, this
matches nothing, because there is no working directory.

This causes commands that depend on LsTree (such as archive,
as used by go in module mode) to fail in bare repos.

### DIFF
--- a/git/lstree.go
+++ b/git/lstree.go
@@ -41,6 +41,9 @@ func LsTree(c *Client, opts LsTreeOptions, tree Treeish, paths []File) ([]*Index
 	// which match the directories of paths, but first check if we're
 	// in the root of the workgir (or using FullTree), in which case we
 	// can just use the tree passed.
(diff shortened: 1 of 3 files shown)


git/lstree.go
```diff
@@ -41,6 +41,9 @@ func LsTree(c *Client, opts LsTreeOptions, tree Treeish, paths []File) ([]*Index
 	// which match the directories of paths, but first check if we're
 	// in the root of the workgir (or using FullTree), in which case we
 	// can just use the tree passed.
+	if c.GetConfig("core.bare") == "true" {
```
sirnewton01: Can this use the IsBare() method defined above?

driusan: done

PR #213, opened 2019-03-06
Add non-head and non-tag ref names in git log

sirnewton01:


git/sha1.go
```diff
@@ -499,8 +499,7 @@ func (c CommitID) GetAllObjectsExcept(cl *Client, excludeList map[Sha1]struct{})
 }
 
 func (c CommitID) getRefNamesList(cl *Client) (string, error) {
-	// FIXME this needs to include other refs that aren't heads or tags
-	refs, err := ShowRef(cl, ShowRefOptions{Heads: true, Tags: true}, []string{})
+	refs, err := ShowRef(cl, ShowRefOptions{Heads: false, Tags: false}, []string{})
```
driusan: This looks like you explicitly don't want them.. why not just `ShowRefOptions{}` to make it more clear that you want the default?

PR #214, opened 2019-03-06
Fetch a remote commit

sirnewton01:
This PR adds the ability to fetch a remote commit. Also, there are now module-specific "go get" tests to help and detect any regressions.


driusan:
Looks fine, but why not do the same thing in the default case of the switch to handle v1 too?

sirnewton01:
> Looks fine, but why not do the same thing in the default case of the switch to handle v1 too?

I was going to do that but v1 seems to require the refname in brackets. I'm not sure what refname to use for an arbitrary commit. What if that commit doesn't have any remote refs pointing to it?

driusan:
v1 doesn't need the ref name in brackets.. it's just in the log for debugging. The Fprint right underneath the log.Print doesn't include it.

sirnewton01:
> v1 doesn't need the ref name in brackets.. it's just in the log for debugging. The Fprint right underneath the log.Print doesn't include it.

Doh! Thanks. I'll fix up V1 in the same way then.

go-get-modules-tests.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "Adding dgit to the path"
+go build
+mkdir -p bin
+cp dgit bin/git
+export PATH=$(pwd)/bin:$PATH
+
+export DGIT_TRACE=/tmp/go-get-dgit-log.$$.txt
+
+# Use a fresh GOPATH
+mkdir /tmp/gopath.$$
```
driusan: Why not unset GOPATH to really make sure it works in module mode?

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: go
 
 go:
   - "1.10.x"
+  - "1.11.x"
```
driusan: Since its released,  can you add 1.12 too?

go-get-modules-tests.sh
sirnewton01: I wanted to make sure that the modules get loaded into a fresh GOPATH separate from the one use to build dgit in the first place. The 'tests' check that those directories get created.

I wonder if GOPATH is going away with the modules or if it is being kept around. The documentation doesn't seem to say one way or another. I suspect that they'll need to keep it around for the module caches.

.travis.yml
sirnewton01: Sure, will do.

go-get-modules-tests.sh
driusan: You're definitely supposed to be able to build outside of $GOPATH in module mode, which is why I'd be more confident in is not being set than depending on an env which I think is only supposed to be a transitionary thing... if you're just concerned about polluting the real $GOPATH, how about setting it to a tmp dir like you are,  and then running the tests in a different tmp dir outside of both $GOPATHs?

sirnewton01: Ok, that sounds good. I've made the change to the script.

driusan: Great, thanks for adding the test.. I'll merge as soon as it passes Travis

PR #216, opened 2019-03-07
Update README.md

Three7Six:
Remove redundant info about Plan 9 (occurred in two consecutive paragraphs) and formalise contractions.


driusan:
Thanks.. I think the whole README needs a rewrite, but this is still an improvement over what's there.

PR #220, opened 2019-03-08
Fix listing of hierarchical tags and branches

### LINKED ISSUE #218: Hierarchical tags and branches aren't working properly
There are cases where hierarchical branches (e.g a/b and abc/def/ghi) aren't working properly.

You can create the branch: ```dgit branch a/b```

But, when you list the branches it only shows the first segment
```
$ dgit branch -l
   a
 * master
```

It's the same for tags.

```
$ dgit tag d/e/f
$ dgit tag -l
d
```

sirnewton01:
Fixes #218


git/client.go
```diff
@@ -256,15 +256,28 @@ func (c *Client) ResetWorkTree() error {
...
+				return err
+			}
+			if !info.IsDir() {
+				files = append(files, path)
+			}
+			return nil
+		})
 	if err != nil {
-		return nil, err
+		return []Branch{}, nil
```
driusan: Why are you returning an empty slice with no err when there's an error?

git/tag.go
```diff
@@ -28,14 +28,25 @@ func TagList(c *Client, opts TagOptions, patterns []string) ([]string, error) {
...
+				return err
+			}
+			if !info.IsDir() {
+				files = append(files, path)
+			}
+			return nil
+		})
 	if err != nil {
-		return nil, err
+		return []string{}, nil
```
driusan: Here too

PR #223, opened 2019-03-08
Fix slice out of bounds problems reading from the error side band channel in fetch pack

### LINKED ISSUE #222: Slice bounds out of range error while running the t5510-fetch tests
A number of tests are failing with these errors.

```
error: git upload-pack: git-pack-objects died with error.
fatal: git upload-pack: aborting due to possible repository corruption on the remote side.
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/driusan/dgit/git.(*packProtocolReader).Read(0xc0000a5000, 0xc0000a2660, 0xc, 0xc, 0x0, 0x0, 0x0)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/fetchpack.go:449 +0xebe
io.(*teeReader).Read(0xc0000b0f40, 0xc0000a2660, 0xc, 0xc, 0xc, 0xc0000a2660, 0x0)
	/usr/local/go/src/io/io.go:535 +0x55
io.ReadAtLeast(0x14817e0, 0xc0000b0f40, 0xc0000a2660, 0xc, 0xc, 0xc, 0x138ab00, 0x13e0501, 0xc0000a2660)
	/usr/local/go/src/io/io.go:310 +0x88
io.ReadFull(0x14817e0, 0xc0000b0f40, 0xc0000a2660, 0xc, 0xc, 0xc, 0x8, 0xc0000b40a0)
	/usr/local/go/src/io/io.go:329 +0x58
encoding/binary.Read(0x14817e0, 0xc0000b0f40, 0x1486f20, 0x1708c30, 0x13af0c0, 0xc0000a2620, 0xc0000b40a0, 0xc0000c4680)
	/usr/local/go/src/encoding/binary/binary.go:245 +0x2de
github.com/driusan/dgit/git.IndexPack(0xc0000b6300, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/indexpack.go:526 +0x2c2
github.com/driusan/dgit/git.IndexAndCopyPack(0xc0000b6300, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/indexpack.go:727 +0x7a
github.com/driusan/dgit/git.fetchPackDone(0xc0000b6300, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x100000000000000, 0x1488980, 0xc0000a4f80, ...)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/fetchpack.go:290 +0xc9c
github.com/driusan/dgit/git.FetchPack(0xc0000b6300, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x100000000000000, 0x7ffeefbff3a0, 0x2, ...)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/fetchpack.go:63 +0x336
github.com/driusan/dgit/git.Fetch(0xc0000b6300, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x100000000000000, 0x7ffeefbff3a0, ...)
	/Users/cmcgee/go/src/github.com/driusan/dgit/git/fetch.go:23 +0x1b5
github.com/driusan/dgit/cmd.Fetch(0xc0000b6300, 0xc0000a40e0, 0x2, 0x2, 0x0, 0x0)
	/Users/cmcgee/go/src/github.com/driusan/dgit/cmd/fetch.go:48 +0x2c7
main.main()
	/Users/cmcgee/go/src/github.com/driusan/dgit/main.go:225 +0x4715
```

sirnewton01:
This seems to fix #222 but it will need another set of eyes to determine whether it is the correct fix or not.


git/fetchpack.go
```diff
@@ -446,11 +446,13 @@ func (p *packProtocolReader) Read(buf []byte) (int, error) {
 				}
 				goto sidebandRead
 			case sidebandErrChannel:
-				n, err := io.ReadFull(p.conn, buf[:size-5])
+				msgbuf := make([]byte, size)
```
driusan: doesn't getting rid of the -5 mean you're overreading? The subtraction is intended to compensate for the length prefix in the pack protocol that was already read.

It's difficult to say what the appropriate value for "n" to return here is.. I think changing the return to 0 probably makes sense and makes the error sideband channel behave more like the status sideband channel (which goes back and does another read, so we never do anything with the n from its read), because implicitly a Read from a pack protocol reader is trying to read the packfile, but the error sideband channel is for remote errors (which is why we return here, unlike sidebandChannel case). Why is it generating a panic? Is the caller not checking the err return? And are the tests that are failing explicitly trying to test for bad fetches?

sirnewton01: One panic was right here because the buf was smaller than size-5. The next happened when I allocated a buffer that was only size-5 and the panic was in io.ReadFull itself. Using size seems to work. I'm not sure why based on what you're saying.

It's hard to pinpoint what exactly is going on with t5510-fetch because there are a lot of things failing at the moment. I think I identified that the git-upload-pack process aborts because it has decided that the repo is corrupt. I figured that we shouldn't be panicking on our end at least.

driusan: Can you add an if statement to compare size-5 to len(buf) before passing the slice to ReadFull?

sirnewton01: I added back the size-5 for the buffer that is allocated. It seems that I was confused about the panic inside io.ReadFull.

driusan:
no, if it works like this I think it's good.

PR #225, opened 2019-03-09
Do not chdir while doing an init, again

driusan:
This was only done so that tests were run in the right directory,
but tests should be the ones doing the Chdir themselves. Changing
the directory causes paths to be evaluated relative to the wrong
directory when handling user input after an init, such as while
doing a clone.

However, the reset command invoked from of a clone needs to be able
to make paths relative.  As result, we still need to chdir within
clone, but restore it for the caller after we're done.

### DIFF
--- a/git/clone.go
+++ b/git/clone.go
@@ -139,6 +139,9 @@ func Clone(opts CloneOptions, rmt Remote, dst File) error {
 	if err := c.GitDir.WriteFile("logs/HEAD", reflog, 0755); err != nil {
 		return err
 	}
+	if opts.Bare {
+		return nil
+	}
 
 	// Finally, checkout the files. Since it's an initial clone, we just
 	// do a hard reset and don't try to be intelligent about what readtree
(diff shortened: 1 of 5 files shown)


git/clone.go
```diff
@@ -132,5 +143,30 @@ func Clone(opts CloneOptions, rmt Remote, dst File) error {
...
+	if err := os.Chdir(c.WorkDir.String()); err != nil {
+		return err
+	}
+	// filepath.Rel can if workdir is ".", so use the absolute path
+	absdir, err := filepath.Abs(".")
+	if err != nil {
+		return err
+	}
+	c.WorkDir = WorkDir(absdir)
+	c.GitDir = GitDir(filepath.Join(c.WorkDir.String(), ".git"))
```
sirnewton01: I wonder if this would be a problem if there was a bare repo being cloned.

driusan: I don't think it would make a difference with the repo being cloned, I think it might break if --bare is passed to clone, but I doubt that's ever worked

driusan: was easy enough to fix by just shortcircuiting the reset if --bare is specified..

PR #229, opened 2019-03-09
Add git reflog exists subcommand

driusan:
This is the simplest subcommand of the reflog command.

### DIFF
--- a/cmd/reflog.go
+++ b/cmd/reflog.go
@@ -1,0 +1,55 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/driusan/dgit/git"
+)
+
+func Reflog(c *git.Client, args []string) error {
+	// flags := newFlagSet("reflog")
+	subcmd := "show"
+	if len(args) > 0 {
+		switch act := args[0]; act {
+		case "show":
+			// show should be an alias of log -g -abbrev-commit
+			// -pretty=oneline, and just delegate there, but log
+			// -g isn't implemented.
+			subcmd = "show"
+		case "expire":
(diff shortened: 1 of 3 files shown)


cmd/reflog.go
```diff
@@ -0,0 +1,55 @@
...
+		return fmt.Errorf("reflog subcommand %v not implemented", subcmd)
+	case "exists":
+		if len(args) != 2 {
+			return fmt.Errorf("usage: %v reflog exists <ref>", os.Args[0])
+		}
+		if git.ReflogExists(c, git.Refname(args[1])) {
+			os.Exit(0)
+		}
+		os.Exit(1)
+		// without this git complains about no return at the end of
```
sirnewton01: s/git/go/ ?

git/reflog.go
```diff
@@ -0,0 +1,6 @@
+package git
+
+// Returns true if a reflog exists for refname r under client.
+func ReflogExists(c *Client, r Refname) bool {
+	return c.GitDir.File(File(r)).Exists()
```
sirnewton01: I thought that the ref logs are the files that are typically in .git/logs/refs/... that store a log of the changes to the ref. Shouldn't this check if c.GitDir.File(File("logs")).File(File(r)).Exists()?

driusan: oops.. you're right.

cmd/reflog.go
driusan: fixed

git/reflog.go
driusan: should be fixed now

PR #232, opened 2019-03-10
Add support for git tag -d to delete a tag

driusan:


git/tag.go
```diff
@@ -94,3 +97,35 @@ tagger %v
 	}
 	return UpdateRefSpec(c, UpdateRefOptions{}, refspec, comm, "")
 }
+
+func TagDelete(c *Client, opts TagOptions, tagnames []Refname) error {
+	if !opts.Delete {
```
sirnewton01: I wonder why pass the options to this operation. Is there something in the options that could change the behaviour of the TagDelete in the future?

driusan: No, it's just a sanity check and for consistency with other commands. It's not entirely clear if `--ignore-case` should work with delete, but I guess git could add some other options for it in the future..

PR #233, opened 2019-03-10
Find the remote and ref specs to fetch from configuration if note specified in the command

sirnewton01:
This gets us slightly closer to be able to run t5510-fetch tests.


cmd/fetch.go
```diff
@@ -38,7 +38,14 @@ func Fetch(c *git.Client, args []string) error {
...
 	var refspecs []git.RefSpec
 	if flags.NArg() < 1 {
-		repository = "origin" // FIXME origin is the default unless the current branch unless there is an upstream branch configured for the current branch
+		// Find the repository we should be fetching
+		headbranch := c.GetHeadBranch()
+		reponame := c.GetConfig(fmt.Sprintf("branch.%s.remote", headbranch.BranchName()))
+		if reponame == "" {
+			reponame = "origin"
+		}
+		fmt.Printf("Udpating %s\n", reponame)
```
driusan: If this a debug statement or is intended to be printed? (In either case there's a typo)

git/fetch.go
```diff
@@ -16,6 +16,16 @@ func Fetch(c *Client, opts FetchOptions, rmt Remote, refs []RefSpec) error {
 	opts.FetchPackOptions.All = (refs == nil)
 	opts.FetchPackOptions.Verbose = true
 
+	// If none were provided then we check to see if there are any
+	//  configured refspecs for this remote
+	if refs == nil {
+		// FIXME this only handles one value for the configuration while there can be many with different refspecs for this operation
+		cfg := c.GetConfig(fmt.Sprintf("remote.%s.fetch", rmt))
+		if cfg != "" {
+			refs = []RefSpec{RefSpec(cfg)}
```
driusan: I think you need to split cfg first in case there's multiple refspecs in the config.

sirnewton01: Right, that would need to happen before we could get the potential multiple values here.

driusan: Actually I think I was mistaken about how multiple refspecs are set.. I thought they were space delimited, but according to https://git-scm.com/book/en/v2/Git-Internals-The-Refspec it's done wth multiple fetch config options.

PR #238, opened 2019-03-13
Hook up the hash options into the show-ref command

sirnewton01:
The option was already present in the ShowRefOptions, but the command-ling flags were not hooked up.


driusan:
Looks like the print statement on line 42 needs to have support added too?

sirnewton01:
> Looks like the print statement on line 42 needs to have support added too?

Yes, I had assumed that the option was controlling the output via the git/showref module, but it is ignored there and the actual printing happens in the cmd. I've made the necessary change.

PR #239, opened 2019-03-14
Add basic for-each-ref command

sirnewton01:
This is just enough to satisfy some of what the t5516-fetch-pull tests need. It supports the basic format only and commits only. The pattern doesn't support wildcards yet.


cmd/foreachref.go
```diff
@@ -0,0 +1,51 @@
...
+		flags.PrintDefaults()
+	}
+
+	// These flags can be moved out of these lists and below as proper flags as they are implemented
+	for _, sf := range []string{"format"} {
+		flags.Var(newNotimplStringValue(), sf, "Not implemented")
+	}
+
+	flags.Parse(args)
+	refs, err := git.ShowRef(c, git.ShowRefOptions{}, []string{})
```
driusan: I think this part that does the work should be in the git package to be consistent with other commands.

driusan: @sirnewton01 ping.. are you going to get a chance to refactor this, or do you want me to pick it up? I don't think it should be a lot of work.

sirnewton01: I can have a look at doing the refactor at some point. Sorry, just really busy at the moment.

PR #242, opened 2019-03-19
Add basic git fsck command

driusan:
This adds a skeleton of the git fsck command. It passes the first 15 tests (looking for basic object corruption and verifying commits) from the official test suite.

### DIFF
--- a/git/fsck.go
+++ b/git/fsck.go
@@ -1,0 +1,316 @@
+package git
+
+import (
+	"compress/zlib"
+	"crypto/sha1"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"math/big"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+type FsckOptions struct {
+	Unreachable      bool
+	NoDangling       bool
+	Root             bool
+	Tags             bool
+	Cache            bool
+	NoReflogs        bool
+	NoFull           bool
+	ConnectivityOnly bool
+	Strict           bool
+	Verbose          bool
+	LostFound        bool
+	NameObjects      bool
+	NoProgress       bool
+}
+
+// Fsck implements the "git fsck" subcommand. It prints any error encountered to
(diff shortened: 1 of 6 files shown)


git/fsck.go
```diff
@@ -0,0 +1,316 @@
...
+			addErr(fmt.Errorf("%v corrupt or missing", obj))
+			continue
+		}
+
+		switch ty := obj.Type(c); ty {
+		case "commit":
+			if err := verifyCommit(c, opts, CommitID(obj)); err != nil {
+				addErr(fmt.Errorf("error in commit %v: %v", obj, err))
+			}
+		case "tree":
```
sirnewton01: Are tags something that could show up in here too?

git/fsck.go
```diff
@@ -0,0 +1,316 @@
...
+			// "foo 1234" is reported as bad email.
+			if strings.Count(s, ">") == 0 {
+				return fmt.Errorf("missingEmail: invalid %v line - missing email", typ)
+			}
+			return fmt.Errorf("badName: invalid %v line - bad name", typ)
+		}
+		if strings.Count(pieces[1], ">") > 0 {
+			return fmt.Errorf("badName: invalid %v line - bad name", typ)
+		}
+		//fmt.Printf(`"%v" "%v" "%v"`+"\n", pieces[1], pieces[2], pieces[3])
```
sirnewton01: Is it worth logging this?

git/fsck.go
driusan: I think so.. there's a few tag related tests which after the commit ones, so it'll probably come up then. First I need to make it past the t1450.15 test, I think, which keeps having a million other little unrelated things that need to be fixed/supported.

git/fsck.go
driusan: I think it would be too verbose to log.. it was mostly just for figuring out which parts of the regex need to be greedy to pass the fsck tests.

PR #254, opened 2019-08-29
Use vanity import path in go get module test

driusan:
Attempting to `go get github.com/golang/text` instead of the
golang.org module name was causing go to fail. Use the
vanity path instead.


driusan:
Still needs #254 for Travis to pass, but this is getting past where it was previously getting stuck.

PR #259, opened 2020-05-04
"Fix" master branch

driusan:
This provides some fixes for the `go get` tests on the master branch. 

1. `GOPROXY=direct` is set, so that dgit is invoked instead of downloading packages through the Go proxy.
2. Options are parsed for `git remote add` (they aren't implemented, just parsed.)
3. A hack is added so that `--` is thrown away in `git remote add`. Go arbitrarily adds it between the remote name and the remote URL, which results in dgit thinking the the remote URL is "--"
4. A hack is added so that "--" is thrown away at the end of `git checkout`. Go arbitrarily adds it, which causes dgit to think that we're trying to checkout files, not a branch.
5. rev-parse no longer thinks commit names are ambiguous if the commit id is arrived to multiple ways (ie. through a pack file and a loose object.)
6. The repository used for `go get` tests is changed. Nearly every repo I could find can't be retrieved with GOPROXY=direct, even with the real git client because Google's git servers require authentication even when doing a clone.
7. Ran `go mod vendor` to update the out of sync vendor directory.
8. 1.13, 1.14, and master are added to allowed failures. They invoke `git cat-file --batch` which we don't yet support, but should be implemented in a separate PR, because we should ensure the cat-file tests from the official git client pass when we implement it.


driusan:
@sirnewton01 Do you think you'd have a chance to review this?

Issue #274, opened 2020-09-02
go get fails due to bitbucket


halfwit:
hg support on bitbucket was blasted away, so we should use the mirror
https://github.com/Plan9-Archive/libauth

driusan:
Thanks for reporting/fixing this.. can't test it right now, but should be fixed by your PR

halfwit:
> Thanks for reporting/fixing this.. can't test it right now, but should be fixed by your PR

Yep, no worries. It seems to build and run correctly on 9front.

PR #275, opened 2020-09-02
Use new import

halfwit:
#274 - I can't test because of modules settings, but I can't see why this won't work fine.


driusan:
It looks like you either forgot to run `go mod vendor` or forgot to add the `github.com/Plan9-Archive/libauth` directory.. I think doing that should also fix Travis.

halfwit:
The latter, yeah. Not sure how I missed that.

PR #280, opened 2022-11-16
GitHub Actions

driusan:
Switch from Travis CI (which wasn't running anyways) to GitHub actions.

Tests are currently failing for a variety of reasons.

Go test ./... is failing because the status tests are failing and need to be fixed and a couple tests are skipped.

go-get-module-tests.sh is failing for an known reason and currently skipped.


driusan:
@sirnewton01 I'm trying to get this repo tests working again. I've got the official tests running (and still passing!) on GitHub Actions instead of Travis CI now but I'm not sure why ./go-get-module-tests is failing. Do you want to look into it for old times sake? Seems to be independent of go version, so I think it might be a change in something downstream in one of the repos that the test depends on.

`go test ./...` failing is a known failure that I really should look into.
(`./go-get-tests` also started failing with newer versions of go, so I restricted it to the last known working version..)

